PR #22, opened 2022-11-01
Allow custom locations when generating a synthetic AST

sylvanc:
In your parser, you can specify generator functions for specific AST node types. This is only used during testing, when generating synthetic AST nodes.

### DIFF
--- a/include/trieste/gen.h
+++ b/include/trieste/gen.h
@@ -1,0 +1,20 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+#include "ast.h"
+#include "xoroshiro.h"
+
+#include <regex>
+
+namespace trieste
+{
+  using sv_match = std::match_results<std::string_view::const_iterator>;
+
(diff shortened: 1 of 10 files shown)


include/trieste/gen.h
```diff
@@ -0,0 +1,20 @@
...
+{
+  using sv_match = std::match_results<std::string_view::const_iterator>;
+
+  using Rand = xoroshiro::p128r32;
+  using Seed = uint64_t;
+  using Result = uint32_t;
+
+  using GenLocationF = std::function<std::string(Rand& rnd)>;
+  using GenNodeLocationF = std::function<Location(Rand&, Node)>;
+}
```
matajoh, line 20: Is it worth moving the Gen class from `wf.h` here as well?

matajoh (approved):
Looks good, just one questions/suggestion.

include/trieste/gen.h
sylvanc: No, it's specific to wf.h. This is just the stuff that's shared between wf.h and parse.h.

PR #30, opened 2023-01-06
Partial application

sylvanc:
This adds passes to generate lower-arity functions for partial application, field accessors, and RHS versions of LHS functions.


samples/verona/lang.cc
```diff
@@ -1337,9 +1479,162 @@ namespace verona
...
+         auto tp = _(TypeParams);
+         auto params = _(Params);
+         size_t start_arity = 0;
+         auto end_arity = params->size();
+         auto parent = f->parent()->parent();
+         auto defs = parent->lookdown(id->location());
+         auto tn = parent->at(wf / Class / Ident, wf / TypeTrait / Ident);
+         Node call = (ref->type() == Ref) ? CallLHS : Call;
+
+         for (auto def : defs)
```
mjp41, line 1510: Perhaps comment that this is for finding smaller arity version.   That is if you have definitions for `f(a,b,c)` and `f(a,b,c,d,e)` then only `f(a,b,c,d)` is provided by the five argument version, and the 2 and 1 argument version come from `f(a,b,c)`.

samples/verona/lang.cc
```diff
@@ -1337,9 +1479,162 @@ namespace verona
...
+         {
+           if ((def == f) || (def->type() != Function))
+             continue;
+
+           auto arity = def->at(wf / Function / Params)->size();
+
+           if ((arity >= start_arity) && (arity < end_arity))
+             start_arity = arity + 1;
```
mjp41, line 1518:
```suggestion
           if (arity < end_arity)
             start_arity = max(start_arity, arity + 1);
```

samples/verona/lang.cc
```diff
@@ -1337,9 +1479,162 @@ namespace verona
...
+         // Create a unique anonymous class name for each arity.
+         for (auto arity = start_arity; arity < end_arity; ++arity)
+           names.push_back(Ident ^ _.fresh());
+
+         Node ret = Seq;
+
+         for (auto arity = start_arity; arity < end_arity; ++arity)
+         {
+           // Create an anonymous class for each arity.
+           auto idx = arity - start_arity;
```
mjp41, line 1532:
```suggestion
           auto name = names[arity - start_arity];
```
and then replace `names[idx]` with `name`?

mjp41 (approved):
LGTM

PR #31, opened 2023-01-11
In the WF-check, make sure we have a tree and not a DAG.

sylvanc:


include/trieste/wf.h
```diff
@@ -509,7 +509,23 @@ namespace trieste
...
+          {
+            out << child->location().origin_linecol()
+                << "this node appears in the AST multiple times:" << std::endl
+                << child->location().str() << child->str() << std::endl
+                << node->location().origin_linecol()
+                << "here:" << std::endl
+                << node->str() << std::endl
+                << child->parent()->location().origin_linecol()
+                << "and here:" << std::endl
+                << child->parent()->str() << std::endl;
```
mjp41, line 523:
```suggestion
                << child->parent()->str() << std::endl
                << "Implementation needs to explicitly clone nodes if they are duplicated.";
```

mjp41 (approved):
LGTM

include/trieste/wf.h
mjp41: Sorry, I missed the new line (std::endl).

PR #32, opened 2023-01-13
Improved error handling in WF and parsing.

sylvanc:
You can now add a `done` callback to your parser to handle terminating sequence-like nodes. You can also generate nicer error messages directly from the parse pass. If an AST contains Error nodes after a pass, future passes aren't run, and if an AST node contains an Error node, WF-check stops checking the parent node.


mjp41 (approved):
LGTM for the core Trieste parts, I didn't review the LLVM/Verona bits

PR #34, opened 2023-01-27
switch to RE2

sylvanc:
This switches from the `stl` regex implementation to `RE2`. While we're no longer able to use predicates or back-references, regex parsing is significantly faster.

### DIFF
--- a/include/trieste/regex.h
+++ b/include/trieste/regex.h
@@ -42,7 +42,12 @@ namespace trieste
         locations.resize(matches);
 
       auto matched = regex.Match(
-        sp, 0, sp.length(), re2::RE2::ANCHOR_START, match.data(), matches);
+        sp,
+        0,
+        sp.length(),
+        re2::RE2::ANCHOR_START,
+        match.data(),
+        static_cast<int>(matches));
 
       if (!matched || (match.at(0).size() == 0))
       {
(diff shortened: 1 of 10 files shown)


include/trieste/regex.h
```diff
@@ -0,0 +1,100 @@
...
+#include "source.h"
+
+#include <re2/re2.h>
+
+namespace trieste
+{
+  class REMatch
+  {
+  private:
+    Source source;
```
mjp41, line 14: Okay, so this is my understanding.  

* `source` is the full underlying string
* `sp` is a view onto the string.  The end of `sp` and the end of `source` are the same.

I wonder if this structure is really an `iterator` rather than `match` as it seems the start of `sp` is the current iteration point into `source`. 

It feels that `skip`, `consume` and `empty` are on the iterator API.

The fields `match`, `locations` and `matches` feel to me like a different type that is the return of `consume`.  Now you don't want to keep allocating them, but I almost wonder if 
```C++
class REIterator
{
   ...
   bool consume(const RE2& regex, REMatch& match)
   ...
};
```
would be a clearer interface.  As this means you don't need to keep doing the allocation, but you get a sense of the correct order of the API.

Probably not worth the change, but it is how I understood the code.

PR #41, opened 2023-05-08
Fixing a bug in the well-formed check

matajoh:
I'm reasonably certain the current logic means that all well-formed checks will pass, as the final return value ignores the `ok` variable.


mjp41 (approved):
LGTM

PR #45, opened 2023-06-19
Adding prefix to fresh names.

matajoh:
It is often advantageous to create a unique variable with a recognisable name to aid in debugging. This PR adds an optional prefix to the `fresh()` call to make this easier.


sylvanc (approved):
LGTM

PR #48, opened 2023-06-21
Exporting CMake dependencies.

matajoh:
This minor change to the CMake configuration makes it so that `trieste` exports the targets of its dependencies in addition to its own, as well as making the dependency relationship explicit. The result of this is that projects which use `trieste` no longer need to explicitly link its dependencies, but rather can just link `trieste::trieste`, like so:

```cmake
target_link_libraries(myproject PUBLIC trieste::trieste)
```


sylvanc (approved):
Nice!

PR #54, opened 2023-07-15
Attempt to allow `const Node&` for creating `WFLookup` objects

matajoh:
This is my attempt at adding this feature. It addresses the problem at the cost of adding the index to the `struct`, thus allowing the `lookup_replace` call to still have access to the correct in-place `Node&` reference.


sylvanc (approved):
`WFLookup` should only ever be on the stack anyway, so I think this is a good, cheap-enough solution.

PR #56, opened 2023-08-21
Disable override on Windows.

mjp41:


mjp41:
@matajoh https://github.com/mjp41/Trieste/tree/backport can be used to test without moving rego forward.

I'll commit some checks into snmalloc for this, so it is not used incorrectly going forward.

sylvanc (approved):

PR #64, opened 2023-09-11
Convert `apply` to be iterative.

mjp41:
This converts the `PassDef::apply` to be iterative rather than recursive.


mjp41:
This is showing a 3x improvement on throughput for Rego. I think we should take it.  Assuming Verona can be made to work with it.

PR #65, opened 2023-09-15
Issue with `Node&` allowing parent circumvention

matajoh:
This PR is about closing off a fairly damaging erroneous usage of the Trieste API. As currently written, client code can write:

```cpp
node->back() = new_node;
```

Because `back()` is a `Node&`, this call will actually change the last child of the node, but circumvent the setting of the parent. The result is that the previous child still points to `node` as its parent, and the new child has a `nullptr` parent. If `node` is then subsequently deleted, the old child still has a pointer to it and this can result in a use-after-free. The new approach is to close this functionality by making these `const Node&` and also changing the iterator to be a const iterator.

To help ease issues where this functionality was needed, I've also added a `replace_at` method which provides a way of replacing a node at an index while maintaining the parent logic.

### DIFF
--- a/include/trieste/ast.h
+++ b/include/trieste/ast.h
@@ -28,7 +28,7 @@ namespace trieste
   }
 
   using Nodes = std::vector<Node>;
-  using NodeIt = Nodes::iterator;
+  using NodeIt = Nodes::const_iterator;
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
LGTM

PR #67, opened 2023-09-16
Adding optional usage of C++ 17

matajoh:
This PR makes a small set of changes which enable the library to be built in C++ 17 with extremely strict flags. No functionality is affected and the C++ 17 path is entirely optional.

### DIFF
--- a/include/trieste/wf.h
+++ b/include/trieste/wf.h
@@ -81,10 +81,26 @@ namespace trieste
           tokens.end(),
           std::back_inserter(offsets),
           [&](const Token& t) {
-            return 1.0 /
-              (1.0 +
-               (alpha * (depth - target_depth) *
-                token_terminal_distance.at(t)));
+            
+            if (token_terminal_distance.find(t) != token_terminal_distance.end())
+            {
+              unsigned long distance = token_terminal_distance.at(t);
+              return 1.0 /
+                (1.0 +
+                (alpha * (depth - target_depth) *
+                  distance));
+            }else{
+              std::ostringstream err;
+              err << "Token " << t.str() << " not found in token_terminal_distance map" << std::endl;
+              err << "{";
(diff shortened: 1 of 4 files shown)


include/trieste/wf.h
```diff
@@ -81,10 +81,26 @@ namespace trieste
           tokens.end(),
           std::back_inserter(offsets),
           [&](const Token& t) {
-            return 1.0 /
-              (1.0 +
-               (alpha * (depth - target_depth) *
-                token_terminal_distance.at(t)));
+            
+            if (token_terminal_distance.find(t) != token_terminal_distance.end())
```
mjp41, line 85: Was this necessary for C++17, or just helpful for debugging.

mjp41 (approved):
LGTM, should we add a CI build that tests C++17 build? Otherwise we'll break it. Also perhaps add a GCC one for the previous PR?

sylvanc (approved):

include/trieste/wf.h
matajoh: I added it in whilst debugging, and decided to keep it as it is otherwise a hard thing to diagnose.

matajoh:
> LGTM, should we add a CI build that tests C++17 build? Otherwise we'll break it. Also perhaps add a GCC one for the previous PR?

I've added them in, and am seeing some build errors for macos-latest + CXX17. Would you be able to take a look?

mjp41:
> > LGTM, should we add a CI build that tests C++17 build? Otherwise we'll break it. Also perhaps add a GCC one for the previous PR?
> 
> I've added them in, and am seeing some build errors for macos-latest + CXX17. Would you be able to take a look?

Should be fixed in upstream snmalloc:
https://github.com/microsoft/snmalloc/pull/635/

PR #71, opened 2023-09-19
Codegen quality: match_children

mjp41:
Adjust code to improve codegen.

### DIFF
--- a/include/trieste/pass.h
+++ b/include/trieste/pass.h
@@ -213,11 +213,6 @@ namespace trieste
           // forwards to next term.
           it += replaced;
         }
-        else
-        {
-          // If we did something, reexamine from the beginning.
-          it = node->begin();
-        }
       }
 
       return changes;
--- a/include/trieste/pass.h
+++ b/include/trieste/pass.h
@@ -120,6 +120,55 @@ namespace trieste
       return (direction_ & f) != 0;
     }
 
+    SNMALLOC_SLOW_PATH ptrdiff_t replace(
+      Match& match,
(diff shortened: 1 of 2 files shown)


include/trieste/pass.h
```diff
@@ -139,74 +189,35 @@ namespace trieste
...
           // forwards to next term.
-          it += std::abs(replaced);
+          it += replaced;
         }
-        else if (replaced >= 0)
+        else
         {
           // If we did something, reexamine from the beginning.
           it = node->begin();
         }
```
mjp41, line 220:
```suggestion
```
Removing this branch, which means that it will retry on the thing it has inserted but not back up to the start of the children results in a 7% speedup in rego-cpp.

This is a semantic change, which means patterns like
```
  A * B => C
  D * C => E
  C * F => G
```
applied to 
```
  D A B F
```
would result in
```
  D A B F -> D C F -> D G
```
rather than
```
  D A B F -> D C F -> E F
```

This change does not appear to affect the results of rego-cpp. Not sure if we should add it.

include/trieste/pass.h
```diff
@@ -139,73 +189,29 @@ namespace trieste
...
+          // If we didn't do anything, advance to the next node.
+          ++it;
         }
-        else
+        else if (flag(dir::once))
         {
-          // Advance to the next node.
-          ++it;
+          // Skip over everything we populated, or if no change (-1) skip
+          // forwards to next term.
```
mjp41, line 213:
```suggestion
          // Skip over everything we populated.
```

sylvanc (approved):

PR #74, opened 2023-09-21
Optimisations again

mjp41:
Two main optimisations:
* Build a fast pattern that allows for quickly deciding which rules to try to apply
* Consolidate T(A) / T(B) into T(A, B) when constructing rules.

### DIFF
--- a/include/trieste/pass.h
+++ b/include/trieste/pass.h
@@ -14,6 +14,88 @@ namespace trieste
     constexpr flag once = 1 << 2;
   }
 
+  /**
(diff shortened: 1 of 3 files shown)


mjp41:
The allow In at the End commit enables
```
T(A) * In(B)
T(A) * In(C)++
```
even when A is the last child. Before this commit, if A was the last child this pattern wouldn't match. 

This commit passes around `parent` rather than `end`. The `parent` can always determine `end`, but the current iterator can only determine the parent before it reaches `end`.

PR #86, opened 2023-11-08
Minor

mjp41:


include/trieste/ast.h
```diff
@@ -656,27 +656,22 @@ namespace trieste
...
-        return std::move(errors);
+        return;
       }
       
       for (auto& child : children)
-        errors = child->get_errors(std::move(errors));
+        child->get_errors(errors);
 
-      return std::move(errors);
+      return;
```
sylvanc, line 674: Unneeded return statement

include/trieste/passes.h
```diff
@@ -163,7 +163,10 @@ namespace trieste
       ok = ok && wf.build_st(ast);
       ok = ok && (!check_well_formed || wf.check(ast));
 
-      auto errors = ast->get_errors();
+      std::vector<Node> errors;
```
sylvanc, line 166: Prefer the type alias `Nodes`

include/trieste/passes.h
```diff
@@ -186,6 +189,9 @@ namespace trieste
       // Check ast is well-formed before starting.
       auto ok = validate(ast, passes);
 
+      PassStatistics stats{};
```
sylvanc, line 192: Prefer removing the extraneous `{}`

include/trieste/regex.h
```diff
@@ -36,7 +36,7 @@ namespace trieste
         match.data(),
         static_cast<int>(matches));
 
-      if (!matched || (match.at(0).size() == 0))
+      if (!matched)
```
sylvanc, line 39: I don't think this is safe, still need to check for a zero length match?

sylvanc (changes requested):

include/trieste/regex.h
sylvanc: Nevermind, this is fine and good

PR #92, opened 2024-03-12
Adding a JSON parser based on Trieste

matajoh:
This code acts as both additional sample code and also as a fully-functional JSON parser and emitter. In addition to the standard Trieste tests it also includes a test driver which runs tests from https://github.com/nst/JSONTestSuite to insure full RFC 8259 compliance.

Further, this PR adds a UTF-8 header to Trieste which provides a range of useful Unicode parsing and conversion functions.

### DIFF
--- a/parsers/json/emitter.cc
+++ b/parsers/json/emitter.cc
@@ -1,0 +1,134 @@
+#include "json.h"
+
+namespace trieste::json
+{
+  JSONEmitter::JSONEmitter(bool prettyprint, const std::string& indent)
+  : m_prettyprint(prettyprint), m_indent(indent)
+  {}
+
+  void JSONEmitter::emit_object(
+    std::ostream& os, const std::string& indent, const Node& object)
+  {
+    if (object->empty())
+    {
+      os << "{}";
+      return;
+    }
+
+    std::string new_indent = indent + "  ";
+    os << "{";
+    if (m_prettyprint)
+    {
+      os << std::endl;
+    }
+    for (std::size_t i = 0; i < object->size(); ++i)
+    {
+      Node member = object->at(i);
+      if (m_prettyprint)
+      {
+        os << new_indent;
+      }
+
+      emit_value(os, new_indent, member->front());
+      os << ":";
+      if (m_prettyprint)
+      {
+        os << " ";
+      }
+      emit_value(os, new_indent, member->back());
+
+      if (i < object->size() - 1)
+      {
+        os << ",";
+      }
+      if (m_prettyprint)
+      {
+        os << std::endl;
+      }
+    }
+    if (m_prettyprint)
+    {
+      os << indent;
+    }
+    os << "}";
+  }
+
+  void JSONEmitter::emit_array(
+    std::ostream& os, const std::string& indent, const Node& array)
+  {
+    if (array->empty())
+    {
+      os << "[]";
+      return;
+    }
+
+    std::string new_indent = indent + "  ";
+    os << "[";
+    if (m_prettyprint)
+    {
+      os << std::endl;
+    }
+    for (std::size_t i = 0; i < array->size(); ++i)
+    {
+      Node element = array->at(i);
+      if (m_prettyprint)
+      {
+        os << new_indent;
+      }
+      emit_value(os, new_indent, element);
+
+      if (i < array->size() - 1)
+      {
+        os << ",";
+      }
+
+      if (m_prettyprint)
+      {
+        os << std::endl;
+      }
+    }
+    if (m_prettyprint)
+    {
+      os << indent;
+    }
+    os << "]";
+  }
+
+  void JSONEmitter::emit_value(
+    std::ostream& os, const std::string& indent, const Node& value)
+  {
+    if (value->in({Number, String, True, False, Null}))
+    {
+      os << value->location().view();
+    }
+    else if (value == Object)
+    {
+      emit_object(os, indent, value);
+    }
+    else if (value == Array)
+    {
+      emit_array(os, indent, value);
+    }
+    else
+    {
+      std::ostringstream message;
+      message << "Unspected node type: " << value->type().str();
+      throw std::runtime_error(message.str());
+    }
+  }
+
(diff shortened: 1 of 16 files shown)


parsers/json/reader.cc
```diff
@@ -0,0 +1,149 @@
...
+    return nullptr;
+  }
+
+  void JSONReader::read()
+  {
+    auto ast = NodeDef::create(Top);
+    Parse parse = json::parser();
+    ast << parse.sub_parse("json", File, m_source);
+    Node maybe_error = has_parse_error(ast);
+    if (has_parse_error(ast))
```
mjp41, line 81: You're calling `has_parse_error` twice.  Is this correct?

parsers/json/reader.cc
```diff
@@ -0,0 +1,149 @@
...
+    m_debug_path = path;
+    return *this;
+  }
+
+  const std::filesystem::path& JSONReader::debug_path() const
+  {
+    return m_debug_path;
+  }
+
+  Node has_parse_error(Node n)
```
mjp41, line 54: Should this be `n->get_errors()`, and we fix that to be iterative?

parsers/json/emitter.cc
```diff
@@ -0,0 +1,134 @@
...
+      emit_object(os, indent, value);
+    }
+    else if (value == Array)
+    {
+      emit_array(os, indent, value);
+    }
+    else
+    {
+      std::ostringstream message;
+      message << "Unspected node type: " << value->type().str();
```
mjp41, line 115:
```suggestion
      message << "Unexpected node type: " << value->type().str();
```

parsers/json/emitter.cc
```diff
@@ -0,0 +1,134 @@
...
+
+    std::string new_indent = indent + "  ";
+    os << "{";
+    if (m_prettyprint)
+    {
+      os << std::endl;
+    }
+    for (std::size_t i = 0; i < object->size(); ++i)
+    {
+      Node member = object->at(i);
```
mjp41, line 26: Should this assert it is a member?

parsers/json/json.cc
```diff
@@ -0,0 +1,108 @@
+#include "json.h"
+
+namespace
+{
+  using namespace trieste;
+  using namespace trieste::json;
+
+  std::size_t invalid_tokens(
+    Node n, std::initializer_list<Token> tokens, const std::string& message)
+  {
```
mjp41, line 10: Given how you use this, would it make sense for it to be 
```C++
std::initializer_list<std::pair<Token,const std::string>> token_messages
```
Then there would only be a single pass, if you are trying to check for several tokens.

parsers/json/json.cc
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    return changes;
+  }
+}
+
+namespace trieste::json
+{
+
+  const auto ValueTokens = T(Object, Array, String, Number, True, False, Null);
```
mjp41, line 32: This only matches one token, so should it be:
```suggestion
  const auto ValueToken = T(Object, Array, String, Number, True, False, Null);
```

mjp41 (changes requested):
Looks good. 

So two high-level things

* I would prefer push/pop versus push/term. I think this would catch errors using the parser framework, and be a bit cleaner. 
* I would be tempted to use the seq operator from the parser, for both `,` and `:`, and this would lead to less work later.

parsers/json/parse.cc
```diff
@@ -0,0 +1,89 @@
...
+         },
+
+       "}" >>
+         [stack](auto& m) {
+           if (stack->empty() || stack->back() != '{')
+           {
+             m.error("Mismatched braces");
+             return;
+           }
+           stack->pop_back();
```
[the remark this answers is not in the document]
matajoh: I've made this change, basically to:

```cpp
m.term()
m.pop(Object)
```

parsers/json/reader.cc
matajoh: Nope that's a bug.

parsers/json/parse.cc
```diff
@@ -0,0 +1,89 @@
...
+           stack->push_back('[');
+         },
+
+       "]" >>
+         [stack](auto& m) {
+           if (stack->empty() || stack->back() != '[')
+           {
+             m.error("Mismatched brackets");
+             return;
+           }
```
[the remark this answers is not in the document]
matajoh: Yup, changed to:

```
m.term()
m.pop(Array)
```

parsers/json/parse.cc
```diff
@@ -0,0 +1,89 @@
...
+         },
+
+       "true" >> [](auto& m) { m.add(True); },
+
+       "false" >> [](auto& m) { m.add(False); },
+
+       "null" >> [](auto& m) { m.add(Null); },
+
+       R"(-?(?:0|[1-9][0-9]*)(?:\.[0-9]+)?(?:[eE][-+]?[0-9]+)?)" >>
+         [](auto& m) { m.add(Number); },
```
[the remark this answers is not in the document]
matajoh: Yeah fair.

parsers/json/json.cc
matajoh: yeah that's a great suggestion

parsers/json/reader.cc
matajoh: I've removed it, and made `n->get_errors()` iterative.

parsers/json/emitter.cc
matajoh: Fixed.

parsers/json/json.cc
matajoh: Changed.

parsers/json/json.cc
matajoh: Fixed.

parsers/json/emitter.cc
matajoh: Agreed, done.

mjp41 (approved):
LGTM

PR #93, opened 2024-03-12
YAML 1.2 parser

matajoh:
This PR adds a YAML 1.2 parser. In addition to the standard Trieste testing this also includes a test driver that runs the official YAML 1.2 test suite provided at https://github.com/yaml/yaml-test-suite.


parsers/yaml/parse.cc
```diff
@@ -0,0 +1,851 @@
...
+          },
+
+        R"((---)([ \t]+))" >>
+          [](auto& m) {
+            m.term();
+            m.pop(Document);
+            m.push(Document);
+            m.add(DocumentStart, 1);
+            m.add(Whitespace, 2);
+          },
```
mjp41, line 286: I wonder if this would be clearer as 
```
        R"((---)([ \t]*))(\r?\n))" >>
          [](auto& m) {
            m.term();
            m.pop(Document);
            m.push(Document);
            m.add(DocumentStart, 1);
            if (m.match(2).len > 0)
              m.add(Whitespace, 2);
            m.add(NewLine, 3);
          },
```
Can you have comments on the start a document line?

parsers/yaml/yaml.cc
```diff
@@ -0,0 +1,2776 @@
...
+
+        In(BlockStart) * BasicToken[Value] >> [](Match& _) -> Node {
+          return err(_(Value), "Invalid text after block scalar indicator");
+        },
+
+        In(SequenceItem) *
+            (T(ValueGroup) << (T(FlowMapping, FlowSequence))[Flow])(
+              [](auto& n) {
+                Node group = n.first[0];
+                Node item = group->parent()->shared_from_this();
```
mjp41, line 1875: I think Trieste should suppport a parent on NodeRange.  This seems to be a common pattern in your code.

parsers/yaml/yaml.cc
```diff
@@ -0,0 +1,2776 @@
...
+          [](Match& _) {
+            return err(_(Value), "Plain dashes in flow sequence");
+          },
+
+        In(FlowMapping, FlowSequence) *
+            (FlowToken[Lhs] * T(Comment)++ * T(Value)[Rhs])([](auto& n) {
+              Node rhs = *(n.second - 1);
+              Location loc = rhs->location();
+              return loc.view().front() == ':';
+            }) >>
```
mjp41, line 927: I think you can use the regexp version of T here:
```suggestion
            (FlowToken[Lhs] * T(Comment)++ * T(Value, ":.*")[Rhs]) >>
```

parsers/yaml/yaml.cc
```diff
@@ -0,0 +1,2776 @@
...
+      dir::bottomup,
+      {
+        In(FlowMapping, FlowSequence) * T(FlowGroup)[FlowGroup] >>
+          [](Match& _) { return Seq << *_[FlowGroup]; },
+
+        In(FlowSequence) *
+            (T(Value)[Value] * T(Comma, FlowSequenceEnd))([](auto& n) {
+              Location loc = n.first[0]->location();
+              return loc.view() == "-";
+            }) >>
```
mjp41, line 917: I think the regex version of T might be useful here:
```suggestion
            (T(Value, "-")[Value] * T(Comma, FlowSequenceEnd)) >>
```

parsers/yaml/yaml.cc
```diff
@@ -0,0 +1,2776 @@
...
+            std::size_t end = view.find_last_not_of(" \t\r\n");
+            loc.pos += start;
+            loc.len = end - start + 1;
+            return Anchor ^ loc;
+          },
+
+        T(Alias)[Alias]([](auto& n) {
+          Node anchor = n.first[0];
+          return anchor->location().view().front() == '*';
+        }) >>
```
mjp41, line 2707:
```suggestion
        T(Alias, "\\*.*")[Alias] >>
```

parsers/yaml/yaml.cc
```diff
@@ -0,0 +1,2776 @@
...
+            return Seq << _(TagPrefix) << (TagName ^ _(TagName));
+          },
+
+        In(TagValue) *
+            (T(TagPrefix)[TagPrefix] * T(TagName)[TagName] *
+             T(Null))([](auto& n) {
+              auto pre = n.first[0]->location().view();
+              auto tag = n.first[1]->location().view();
+              return pre == "!!" && tag == "str";
+            }) >>
```
mjp41, line 2433:
```suggestion
            (T(TagPrefix, "!!")[TagPrefix] * T(TagName, "str")[TagName] *
             T(Null)) >>
```

mjp41 (commented):
Here are a bunch of drive by comments.  I think the regexp thing will tidy up the code a little, but I haven't tried to use it.

I quite like the use of actions to handle a lot of the corner cases in the parsing for indents and new lines.  It was quite readable. I don't know if they were correct though ;)

parsers/yaml/yaml.cc
```diff
@@ -0,0 +1,2776 @@
...
+            (T(BlockLine)[Lhs] * T(BlockLine)[Rhs])([](auto& n) {
+              return n.first[0]->location().len == 0 &&
+                n.first[1]->location().len == 0;
+            }) >>
+          [](Match& _) { return _(Lhs); },
+
+        In(SingleQuote, DoubleQuote) *
+            (T(EmptyLine)[Lhs] * T(BlockLine)[Rhs])(
+              [](auto& n) { return n.first[1]->location().len == 0; }) >>
+          [](Match& _) { return _(Lhs); },
```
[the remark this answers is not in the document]
matajoh: I can't do this operation until after the quotes pass has completed, and writing it as a post meant that I had to write all the matching logic by hand, and it seemed silly to add a pass just to do this clean-up. What would you propose instead?

parsers/include/trieste/yaml.h
```diff
@@ -0,0 +1,488 @@
...
+    wf_tags
+    | (SingleQuote <<= (BlockLine|EmptyLine)++[1])
+    | (DoubleQuote <<= (BlockLine|EmptyLine)++[1])
+    | (Literal <<= AbsoluteIndent * ChompIndicator * Lines)
+    | (Folded <<= AbsoluteIndent * ChompIndicator * Lines)
+    | (Lines <<= (BlockLine|EmptyLine)++)
+    ;
+  // clang-format on
+
+  inline const auto wf_anchors = wf_quotes;
```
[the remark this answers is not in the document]
matajoh: Fixed in upcoming PR

parsers/yaml/yaml.cc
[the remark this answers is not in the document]
mjp41: A comment to explain.  Otherwise, I am trying to understand why changing these empty lines affects anchors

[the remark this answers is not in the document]
matajoh: Yup, great point

PR #94, opened 2024-03-14
Fix recursive stack depth problems.

mjp41:
The rewrite engine used an efficient path based traversal that was iterative with effectively pre and post-order actions.  This PR factors that into ast.h, and then uses it for various operations to prevent stack overflow on deep asts.

It also fixes the recursive delete problem that you get from shared_ptr, by building a work list of nodes that need to be destructed.

### DIFF
--- a/include/trieste/ast.h
+++ b/include/trieste/ast.h
@@ -680,6 +680,36 @@ namespace trieste
       out << ")";
     }
 
+    template<typename Pre, typename Post>
+    SNMALLOC_FAST_PATH void traverse(Pre pre, Post post)
+    {
+      Node root = shared_from_this();
+      if (!pre(root))
+        return;
+
+      std::vector<std::pair<Node&, NodeIt>> path;
+      path.push_back({root, root->begin()});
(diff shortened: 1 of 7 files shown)


matajoh (approved):
I love the way you've approached this. Nice. No notes.

include/trieste/ast.h
```diff
@@ -155,7 +155,37 @@ namespace trieste
...
+    {
+      // Contains a pointer to a vector on the stack for handling the
+      // recursive destruction of the AST.
+      // We don't use an actual vector in the TLS as the destruction
+      // of the TLS can cause problems on some platforms.
+      thread_local std::vector<Nodes>* work_list;
+
+      if (work_list)
+      {
+        work_list->push_back(std::move(children));
```
matajoh, line 168: Minor suggestion: add a comment here to make it explicit that this will only ever be called by the loop below because the variable is a thread_local. The only circumstance in which this branch is reached is to queue more things to destruct as a result of nodes.clear() on line 184. That was the only stumbling block I had.

PR #98, opened 2024-04-05
Adding Reader, Writer, and Rewriter to Trieste.

matajoh:
In the context of adding a means to convert YAML to JSON, this PR adds some new constructs to Trieste.
- A Reader is a generalized code construct which a language implementer can expose that parses a Source and returns an AST.
- A Writer is the same but for taking an AST and writing it to a Destination (also added in this PR)
- A Rewriter converts one AST to another AST via one or more rewrite passes.

In all three cases the implementer controls the activity of the construct via a series of rewrite passes. As part of this work, Driver was rewritten to optionally use a Reader, and the Process classes was refactored to make it a bit more flexible and easy to use.

Additionally, the YAML language implementation exposes a `to_json` converter, as well as a `event_writer` (for the existing event stream output) and a `writer` which emits YAML. All three are validated against the test suite. Both YAML and JSON expose Reader objects, and JSON exposes a Writer as well (a refactoring of its emitter).


include/trieste/passes.h
```diff
@@ -112,65 +113,178 @@ namespace trieste
     std::chrono::microseconds duration;
   };
 
+  struct ProcessResult
+  {
+    bool ok;
+    std::string last_pass;
+    Node ast;
+    Node error;
```
mjp41, line 121: Would this be better as a `Nodes`?  Then you don't need to introduce an `ErrorSeq` token?

include/trieste/passes.h
```diff
@@ -216,120 +333,95 @@ namespace trieste
...
 
       wf::pop_front();
 
-      return ok;
+      return {ok, last_pass, ast, error};
     }
   };
 
-  inline void print_errors(Nodes& errors, logging::Log& err)
+  class Rewriter
```
mjp41, line 350: Should this be in its own file?

mjp41 (approved):
LGTM.  Could of very minor things.

include/trieste/passes.h
matajoh: not a bad idea

mjp41 (approved):
LGTM. thanks

PR #106, opened 2024-04-18
Prevent rewriting a term that contains an error.

mjp41:


mjp41:
@matajoh @sylvanc not sure if we want this change in semantics.  It would fix 

> https://github.com/microsoft/verona/pulls/#667

without requiring code changes.

It would make it impossible to explicitly match an `Error` node.

matajoh:
I think this is a small price to pay for avoiding various issues caused by inadvertently matching errors.

PR #110, opened 2024-04-26
Added Shrubbery parser as a sample

EliasC:
This PR adds a sample that parses [Shrubbery Notation](https://plt.cs.northwestern.edu/pkg-build/doc/shrubbery/index.html). It complements the `infix` example by having a slightly more involved parsing pass (whereas the three rewriting passes are quite straightforward). In particular, it gives an example of handling indentation sensitivity. There is a README that gives an overview of Shrubbery parsing and the implementation. The source code has explanatory comments, but assumes that the reader has done the `infix` tutorial and therefore understands the basics of parsing and pattern matching.

The implementation relies on #90.


EliasC:

matajoh:
Small question: where do the test files come from? If they have been copied in from another repo, could we potentially clone that repo as part of the build. to avoid having all of these in the repo? Not a big deal if not, just curious. You can see an example of what I'm talking about in the parser directory:

```cmake
if(TRIESTE_BUILD_PARSER_TESTS)
    enable_testing()
    add_subdirectory(test)

    if(NOT IS_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/JSONTestSuite)
    execute_process(COMMAND ${GIT_EXECUTABLE} clone --depth=1 https://github.com/nst/JSONTestSuite
                    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
                    OUTPUT_QUIET)
    endif()

    if(NOT IS_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/yaml-test-suite)
    execute_process(COMMAND ${GIT_EXECUTABLE} clone --depth=1 --branch data-2022-01-17 https://github.com/yaml/yaml-test-suite
                    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
                    OUTPUT_QUIET)
    endif()
endif()
```

It would also be neat for these tests to be added either via a test driver like [json_test.cc](https://github.com/microsoft/Trieste/blob/main/parsers/test/json_test.cc) or by adding them in CMake:

```cmake
add_test(NAME json_trieste COMMAND json_trieste test -f WORKING_DIRECTORY $<TARGET_FILE_DIR:json_trieste>)
```

Anything to have more robust CI :)

samples/shrubbery/parse.cc
```diff
@@ -0,0 +1,332 @@
...
+  // f(1) + 2
+  //   + 3
+  //   - 4
+  //
+  // is identical to `f(1) + 2 + 3 - 4`. The plus on the second line can
+  // arbitrarily indented, but the minus must be in the same column as the plus
+  // on the previous line. When parsing the program above, the Indent will be
+  // ((0,0), 2), indicating that the established indentation level is line 0
+  // colum 0, and that this line is continued on column 2.
+  struct Indent {
```
matajoh, line 32: As a general note on Trieste usage, there are two strategies that may make your life considerably easier as regards meaningful whitespace. The first, and perhaps simplest, is to make the parser even dumber and have it explicitly capture whitespace as `Whitespace` and `NewLine` tokens, i.e.:

```c++
  R"(\r?\n)" >> [](auto& m){
    m.add(NewLine);
  }

  R"([ \t]+)" >> [](auto& m){
    m.add(Whitespace);
  }
```

You can then defer decisions about indents etc. to a dedicated pass in Trieste, where you'll have much more flexibility and context to make your decisions. This works rather well. Another option is to use much the same pattern, but to use pattern matching to measure the indent explicitly while processing the end of the previous line, like so: 

```c++
  R"(\r?\n([ \t]*))" >> [](auto& m){
    size_t indent = m.match(1).len;
    // your indent logic
  }
```

This lookahead can save a bunch of bookkeeping (you can preemptively push or pop indent blocks).

samples/shrubbery/shrubbery.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <trieste/driver.h>
```
matajoh, line 3: Best practice is now to include `trieste/trieste.h`.

samples/shrubbery/shrubbery.h
```diff
@@ -0,0 +1,26 @@
...
+  inline const auto Semi    = TokenDef("semicolon"); // ;
+  inline const auto Alt     = TokenDef("alt");       // |
+  inline const auto Op      = TokenDef("op", flag::print); // Operators
+  inline const auto Atom    = TokenDef("atom", flag::print); // Everything else
+
+  // Used as identifiers
+  inline const auto Id = TokenDef("id");
+  inline const auto Lhs = TokenDef("lhs");
+  inline const auto Rhs = TokenDef("rhs");
+
```
matajoh, line 23: Best practice is to export your final WF definition here, as otherwise people will not be able to easily use the ASTs you produce downstream.

samples/shrubbery/reader.cc
```diff
@@ -0,0 +1,128 @@
...
+    };
+  }
+
+  // Merge alternatives into a single Alt node containing blocks
+  PassDef merge_alternatives()
+  {
+    return {
+      "merge alternatives",
+      wf_alternatives,
+      dir::topdown,
```
matajoh, line 75: I think this could be turned into a `dir::once` pass by adding `++` to the captures below? I'll give an example of what I mean.

samples/shrubbery/reader.cc
```diff
@@ -0,0 +1,128 @@
...
+        // a comma or semi-colon
+        (--In(Comma, Semi) * T(Group)[Group] * (T(Group) << (T(Alt)[Alt]))) >>
+          [](Match& _) { return _(Group) << _(Alt); },
+
+        // Alternatives keep their contents in a block
+        (T(Alt)[Alt] << !T(Block)) >>
+          [](Match& _) { return Alt << (Block << *_[Alt]); },
+
+        // Merge sequence of alternatives into a single alternative
+        (T(Alt)[Alt] * (T(Alt) << T(Block)[Block])) >>
```
matajoh, line 87: Potential alternative to allow `dir::once`?

```c++
(T(Alt)[Alt] * (T(Alt) << T(Block)))++[Group] >>
  [](Match& _) {
    Node alt = _(Alt);
    for(auto& node : _[Group]){
      alt << node->front();
    }
    return alt;
  }
```

matajoh: (NB this requires my PR which changes `NodeRange` to a span)

samples/shrubbery/reader.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  // Remove nodes for commas and semicolons and replace them by their children.
+  // Also check for empty alternatives, which may only appear immediately under
+  // braces and brackets
+  PassDef drop_separators()
+  {
+    return {
+      "drop separators",
+      wf_no_semis_or_commas,
+      dir::topdown,
```
matajoh, line 101: This can be a `dir::bottomup|dir::once` pass.

samples/shrubbery/reader.cc
```diff
@@ -0,0 +1,128 @@
...
+  {
+    return Error << (ErrorMsg ^ msg) << (ErrorAst << node);
+  }
+
+  PassDef check_parsing()
+  {
+    return {
+      "check parsing",
+      wf_check_parser,
+      dir::topdown,
```
matajoh, line 22: This can be a `dir::bottomup|dir::once` pass I believe

matajoh (dismissed):
Most of these comments are non-blocking, i.e., more suggestions for improvements that could be done in the future.

I would like (if possible) to not host the test files in this repo, though.

samples/shrubbery/reader.cc
EliasC: Are `dir::bottomup` passes more efficient than `dir::topdown` passes? I can see how `dir::once` is preferrable.

samples/shrubbery/reader.cc
EliasC: Having foreach loops for `NodeRange` looks very nice! Are the parentheses in the pattern off by one? Reading the code, I would have guessed that the `Group` capture should not include the `Alt` capture, but only the subsequent nodes (so that we can add their blocks under the `alt` node in the effect.

samples/shrubbery/parse.cc
EliasC: I was thinking about a version with the explicit whitespace tokens, but tried writing the Shrubbery parser as a way to stress the parser module more than the existing samples (when this was developed, infix was the only other sample in the repo). It is also nice how, modulo whitespace sensitivity, the Trieste parser sort of looks like a generalized shrubbery parser. The second option does look a lot nicer than what I am doing right now though, so I will give that a spin!

EliasC:
> Small question: where do the test files come from? If they have been copied in from another repo, could we potentially clone that repo as part of the build. to avoid having all of these in the repo?

I wrote them myself. Some of them come from the [documentation of shrubbery notation](https://plt.cs.northwestern.edu/pkg-build/doc/shrubbery/index.html). I have been looking for an existing test suite without success (also there would need to be a filtering process since I don't support all of Shrubbery notation).

EliasC:
@matajoh I have addressed most of your comments now. 
- Every rewrite pass is now `bottomup | once` and generally looks nicer. 
- I removed all the examples for now and will keep them around locally until I figure out a way to add them in a nice way or find an external test suite. They are quite weak as test cases as you still need to manually look at the produced trees to see if they did the right thing.
- I tried experimenting with your suggested solutions for whitespace, but didn't figure out a good to do it. The problem with counting whitespace after newlines is that a line could start with spaces but continue with a comment (or nothing) meaning it shouldn't be counted. Since there is no lookahead in the parsing step, I don't know whether to check indentation levels until I actually see a term that should be added. For adding explicit whitespace tokens, it becomes quite hairy for input like
  ```
  one: two: three: four:
       two again
  ```
  where `two again` should belong to the same block as `two`. It's doable, but requires non-trivial logic that in the end isn't a lot cleaner than what I'm doing now.

I also added a final rewriting pass that takes the tree to a shape that matches the grammar of shrubbery notation (this is presented in `shrubbery.h`). Later it might make sense to add a writer that outputs the parsed program as an S-expression (or indeed as shrubbery notation), but for now I think it works as an example to learn from.

mjp41 (approved):

PR #111, opened 2024-05-07
do not purge install prefix (microsoft#96)

dariuskl:
Removes a piece of CMake code that is responsible for clearing the installation directory.


mjp41 (approved):
LGTM

@sylvanc I think this is from how Verona was using this for building a layout for the standard libraries.  I think we should move any custom code of that nature into Verona.

mjp41:
@dariuskl thanks for your contribution.

Issue #96, opened 2024-03-21
Trieste clears the install directory


dariuskl:
I was debugging a hard-to-track issue in a build of rego where random files would be missing from the install tree after a call to `cmake --install`. Ultimately, it turned out to be the following command in the `CMakeLists.txt` of trieste.

https://github.com/microsoft/Trieste/blob/dd7c8e031f630e47bb895ff77cf0eee7d49dace0/CMakeLists.txt#L110

I think it's not a good idea to do that, in particular in situations where the user install into `/usr/local`. But even for other destinations, trieste should not assume to be the only library in that tree. Is there a particular reason to do that, that I'm missing?

PR #113, opened 2024-05-17
Adding a flag to enable cleaning the install directory

matajoh:
The previous PR #111  removed the install directory cleaning functionality that is expected by various downstream repositories. This PR adds it back in with a flag, to address the concerns raised in #96. Downstream users will need to set `TRIESTE_CLEAN_INSTALL` to `On` to get the old behaviour.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ option(TRIESTE_BUILD_SAMPLES "Specifies whether to build the samples" ON)
 option(TRIESTE_BUILD_PARSERS "Specifies whether to build the parsers" ON)
 option(TRIESTE_BUILD_PARSER_TESTS "Specifies whether to build the parser tests" OFF)
 option(TRIESTE_USE_CXX17 "Specifies whether to target the C++17 standard" OFF)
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
LGTM

Seems like ASAN has been tripped by something (unrelated to this, but should be fixed).

parsers/yaml/reader.cc
```diff
@@ -2440,15 +2440,15 @@ namespace
         In(Plain) * (T(Line)[Line] << Any) >>
           [](Match& _) {
             Node line = _(Line);
-            if (line->front()->type() == Whitespace)
+            if (line->front() == Whitespace)
             {
               line->erase(line->begin(), line->begin() + 1);
             }
-            if (line->back()->type() == Whitespace)
+            if (!line->empty() && line->back() == Whitespace)
```
mjp41, line 2447: How about:
```C++
            if (line->empty())
            {
              return BlockLine ^ line;
            }
            if (line->back() == Whitespace)
            {
              line->pop_back();
            }
            if (line->back() == Comment)
            {
              return err(line, "comment in multiline plain scalar");
            }
```

PR #114, opened 2024-05-30
Adding the `Fuzzer`class, `NodeRange` -> `span`, and WF contexts.

matajoh:
This commit adds a new `Fuzzer` class, which encapsulates the logic from `Driver` and allows it to be applied to the new `Reader`, `Writer`, and `Rewriter` classes. Another global change is the addition of a stack of wellformedness contexts. By calling `wf::new_context` and `wf::end_context`, it is possible to have nested Trieste pass ranges execute. This allows a Trieste pass to execute another Trieste pass as part of its execution (e.g. to parse a string using the JSON parser and the convert the resulting AST to a language-specific format). Finally, this commit changes `NodeRange` to be a `std::span` (NB an equivalent class is provided for C++17).

### YAML
- Fixed a bug with block indents that contain YAML or JSON
- Fixed a bug where a colon is immediately adjacent to the start of a flow construct
- Fixed a bug with colons immediately adjacent to values inside flow constructs
- Fixed an overeager error message for comments that look like mapping keys
- Fixed a bug with bad aliases in `to_json`

### JSON
- Added `unescape` and `escape` functions for handling JSON strings
- Added a symbol table to `Object` and a `Key` token to make JSON ASTs easier to use

### Misc
- Added some code to mitigate situations in which P(c | d, p) is uniform during fuzzing.
- Added `find_first` and `contains` methods on `Node` which implement `Token` specific logic.
- Added `postparse` and `output_wf` properties for `Reader`
- Added `input_wf`, `output_wf` and `passes` properties for `Rewriter`
- Added `input_wf` and `passes` properties for `Writer`

### DIFF
--- a/include/trieste/ast.h
+++ b/include/trieste/ast.h
@@ -473,7 +473,7 @@ namespace trieste
 
     void push_back(NodeRange range)
     {
-      for (Node n : range)
+      for (Node& n : range)
         push_back(n);
     }
 
@@ -488,7 +488,7 @@ namespace trieste
 
     void push_back_ephemeral(NodeRange range)
     {
-      for (Node n : range)
+      for (Node& n : range)
         push_back_ephemeral(n);
     }
 
@@ -683,7 +683,7 @@ namespace trieste
       return parent(Top)->fresh(prefix);
     }
 
-    Node clone()
+    Node clone() const
     {
       // This doesn't preserve the symbol table.
       auto node = create(type_, location_);
@@ -1012,7 +1012,7 @@ namespace trieste
 
   inline std::ostream& operator<<(std::ostream& os, const NodeRange& range)
   {
-    for (Node n : range)
+    for (const Node& n : range)
       n->str(os);
 
     return os;
--- a/include/trieste/ast.h
+++ b/include/trieste/ast.h
@@ -10,6 +10,10 @@
 #include <sstream>
 #include <vector>
 
+#ifndef TRIESTE_USE_CXX17
+#  include <span>
+#endif
+
 namespace trieste
 {
   struct indent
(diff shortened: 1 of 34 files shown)


include/trieste/rewrite.h
```diff
@@ -1119,19 +1119,17 @@ namespace trieste
 
   inline Node operator<<(Node node, detail::RangeContents range_contents)
   {
-    for (auto it = range_contents.range.first;
-         it != range_contents.range.second;
-         ++it)
+    for (Node n : range_contents.range)
```
mjp41, line 1122: To avoid unnecessary RC manipulations can you use a reference?
```suggestion
    for (Node& n : range_contents.range)
```

include/trieste/ast.h
```diff
@@ -391,8 +473,8 @@ namespace trieste
 
     void push_back(NodeRange range)
     {
-      for (auto it = range.first; it != range.second; ++it)
-        push_back(*it);
+      for (Node n : range)
```
mjp41, line 476:
```suggestion
      for (Node& n : range)
```

include/trieste/ast.h
```diff
@@ -930,8 +1012,8 @@ namespace trieste
 
   inline std::ostream& operator<<(std::ostream& os, const NodeRange& range)
   {
-    for (auto it = range.first; it != range.second; ++it)
-      (*it)->str(os);
+    for (Node n : range)
```
mjp41, line 1015:
```suggestion
    for (Node& n : range)
```

include/trieste/rewrite.h
```diff
@@ -1171,10 +1169,10 @@ namespace trieste
   inline Nodes clone(NodeRange range)
   {
     Nodes nodes;
-    nodes.reserve(std::distance(range.first, range.second));
+    nodes.reserve(range.size());
 
-    for (auto it = range.first; it != range.second; ++it)
-      nodes.push_back((*it)->clone());
+    for (Node n : range)
```
mjp41, line 1174:
```suggestion
    for (Node& n : range)
```

mjp41 (approved):
LGTM. Minor couple of Node& instead of Node.  I would prefer Node& as it doesn't require RC manipulations where it is safe to use it.

mjp41:
FYI, I didn't review the json/yaml/infix changes.

Issue #116, opened 2024-06-03
Flag to disable snmalloc


matajoh:
We may want to relax the hard dependency on snmalloc to be gated using a CMake option (that defaults to ON). That would allow downstream consumers of the library to opt out of snmalloc usage if they so desire.

mjp41:
Closed by #117

PR #117, opened 2024-06-06
Make snmalloc optional

mjp41:
This allows snmalloc to be disabled.  It also only overrides new/delete.


mjp41:
Addresses #116

CMakeLists.txt
```diff
@@ -80,11 +76,19 @@ target_include_directories(trieste
 
 target_link_libraries(trieste
   INTERFACE
-  snmallocshim-static
   re2::re2
   CLI11::CLI11
+  snmalloc
```
matajoh, line 81: Just for my own understanding, why does Trieste need to link against snmalloc even if it doesn't use it?

matajoh (approved):

CMakeLists.txt
mjp41: So Trieste uses some of snmallocs headers.  There are a bunch of cross platform defines that are useful.  `SNMALLOC_FAST_PATH` and `SNMALLOC_SLOW_PATH` which I use in `snmalloc` heavily to get the code gen to inline precisely how I need it to.  I use that in the core of the rewrite engine to similarly get the code gen I was looking for.

That is not linking against any binaries with that target, only the header files.

mjp41: Sorry, it was actually in the Logging library, but for code gen purposes.

PR #121, opened 2024-06-14
Fix #120 by commenting out the GIT_SHALLOW TRUE

### LINKED ISSUE #120: CMake can't check out snmalloc on first run
To reproduce:
- clone a fresh copy of Trieste
- `mkdir build && cd builde && cmake ..`
- CMake version 3.28.3, which is the default on Ubuntu 24.04, if that turns out to matter.

The outcome is that Trieste can't checkout its desired commit of snmalloc, and build file generation fails.

The key issue seems to be `GIT_SHALLOW TRUE` in the `FetchContent_Declare` command in the `CMakeLists.txt`.
I found a long and complicated discussion among CMake devs which repeatedly hints that this should never be used with git hashes (https://gitlab.kitware.com/cmake/cmake/-/issues/17770), and a recent commit to main here seems to have done just that.

The simple fix would therefore be to delete the `GIT_SHALLOW` flag and add a comment on this whole situation... or switch to a named branch, but I see why tracking main like it used to would be problematic.

fhackett:
This is the simplest fix for the linked issue I could think of, posted for comment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,11 @@ FetchContent_Declare(
   snmalloc
   GIT_REPOSITORY https://github.com/microsoft/snmalloc
   GIT_TAG 835ab5186325736ff5aa8ef53728970d1d74c3bc
-  GIT_SHALLOW TRUE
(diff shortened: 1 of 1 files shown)


matajoh (approved):
LGTM

PR #122, opened 2024-06-18
Fix for large JSON files

matajoh:
Very large JSON files will have residual stack values (due to the preview check) which were not clear, meaning that they could not successfully parse.

### DIFF
--- a/parsers/json/parse.cc
+++ b/parsers/json/parse.cc
@@ -91,6 +91,7 @@ namespace trieste
         {
           m.error("Mismatched braces or brackets");
         }
+        stack->clear();
       });
 
       return p;


mjp41 (approved):
LGTM.  Will the mismatched brackets part be a problem for this? I.e. if the first chunk has mismatched brackets.

matajoh:
This is why I check error ratios. The issue here was that the check for validity was passing (as the error ratio was very small, just the one mismatched braces error at the end of the file) but then when the valid file was parsed the stack had the residual braces and brackets from the previous parse and would be marked as being in error, though it wasn't. By clearing the stack the problem goes away.

Issue #115, opened 2024-06-03
Large binary sizes


matajoh:
Trieste executables are currently quite large, for example, on Ubuntu in Debug:

```
~/Trieste/build/dist$ tree -sh parsers
[4.0K]  parsers
├── [ 13M]  json_fuzzer
├── [ 13M]  json_test
├── [ 22M]  yamlc
├── [ 21M]  yaml_fuzzer
└── [ 24M]  yaml_test

~/Trieste/build/dist$ tree -sh infix
[4.0K]  infix
├── [ 14M]  infix
└── [ 13M]  infix_trieste
```

Even in Release, they are larger than may be warranted (in particular, for `infix`):

```
~/Trieste/build/dist$ tree -sh parsers
[4.0K]  parsers
├── [1.3M]  json_fuzzer
├── [1.4M]  json_test
├── [3.6M]  yamlc
├── [3.6M]  yaml_fuzzer
└── [3.8M]  yaml_test

~/Trieste/build/dist$ tree -sh infix
[4.0K]  infix
├── [1.5M]  infix
└── [1.4M]  infix_trieste
```

fhackett:
I have investigated this using a mix of disassembly and binary analysis.
There are two short answers, one for debug and one for release builds.
For debug, only 10% of the binary is actually code; all the rest seems to be various metadata and debug info.
The code that is there seems fine, though clearly a lot can be pruned - RE2 seems benefit a lot from dead code elimination in release mode, for example.
For release, most of the binary is code (70% roughly), and the largest functions, at 2-6% of total program symbol bytes, are Trieste rule definitions with lots of rules.
I mean the functions which when called produce `PassDef`, not the associated lambdas that do actual work.
These seem very big for what they are.

## Methodology

Tools used:
- Ghidra CodeBrowser
- elf-size-analyze wrapper around binutils (lists symbols with binary size info)

I tested the [rego-cpp](https://github.com/microsoft/rego-cpp) implementation and my local fork of infix (mostly the same thing, but some code is added that may affect specific sizes of things).
I looked at both release and debug builds of those two programs.
I have attached data dumps I got from using elf-size-analyze.
You can reproduce these by installing [elf-size-analyze](https://pypi.org/project/elf-size-analyze/) and running `elf-size-analyze --rom --no-color <the binary>`, which should produce similar data.
To see mangled symbols which might cross-reference better with Ghidra's output given the same file, add `--no-demangle`.

[...]

PR #124, opened 2024-06-25
Binary size improvements related to #115

fhackett:
See issue #115 for more discussion.

The key idea here is to use the `SNMALLOC_SLOW_PATH` macro, which is a cross-plaform noinline macro, on functions which construct, copy, and destroy long-lived pass, pattern, and WF objects.

These functions are only called on start-up, so they should have minimal performance impact.
By default, they are inlined into Trieste pass defs and such.
Some of their contents is rather bulky, like red-black tree copy routines, big reference count updates (including dynamic checks for whether glibc is multi- or single- threaded), and array resizing code.

Making these lifecycle functions explicit using `= default` and marking them with `SNMALLOC_SLOW_PATH`, alongside some other operators that might possible cause similar problems, saves 15-27% on code size for executables using Trieste.

### DIFF
--- a/include/trieste/wf.h
+++ b/include/trieste/wf.h
@@ -248,9 +248,9 @@ namespace trieste
       Choice choice;
       size_t minlen;
 
-      SNMALLOC_SLOW_PATH Sequence() = default;
       SNMALLOC_SLOW_PATH Sequence(Choice choice_, size_t minlen_) :
         choice{choice_}, minlen{minlen_} {}
+      SNMALLOC_SLOW_PATH Sequence() = default;
       SNMALLOC_SLOW_PATH Sequence(const Sequence&) = default;
       SNMALLOC_SLOW_PATH Sequence(Sequence&&) = default;
       SNMALLOC_SLOW_PATH Sequence& operator=(const Sequence&) = default;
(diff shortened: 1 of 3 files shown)


davidchisnall:
It would probably be nice to define a `TRIESTE_SLOW_PATH` macro, rather than import the snmalloc one (even if it's just `#define TRIESTE_SLOW_PATH SNMALLOC_SLOW_PATH`)

fhackett:
@davidchisnall  I saw a lot of `SNMALLOC_SLOW_PATH` and `SNMALLOC_FAST_PATH` in the codebase already, so the second commit adding that would get somewhat bigger.

That said, I would be fine doing that. It's just a search-replace.

include/trieste/wf.h
```diff
@@ -239,6 +248,15 @@ namespace trieste
       Choice choice;
       size_t minlen;
 
+      SNMALLOC_SLOW_PATH Sequence() = default;
+      SNMALLOC_SLOW_PATH Sequence(Choice choice_, size_t minlen_) :
```
mjp41, line 252: Could you put the non-default one at the top please.

include/trieste/wf.h
```diff
@@ -306,6 +324,15 @@ namespace trieste
       std::vector<Field> fields;
       Token binding;
 
+      SNMALLOC_SLOW_PATH Fields() = default;
+      SNMALLOC_SLOW_PATH Fields(std::vector<Field> fields_, Token binding_) :
```
mjp41, line 328: Same

include/trieste/wf.h
```diff
@@ -455,6 +482,15 @@ namespace trieste
       Token type;
       ShapeT shape;
 
+      SNMALLOC_SLOW_PATH Shape() = default;
+      SNMALLOC_SLOW_PATH Shape(Token type_, ShapeT shape_) :
```
mjp41, line 486: Same

mjp41 (approved):
LGTM.  I makes me wonder if we should really have the C++ operator overloads generate an AST, and then build the pattern matching structure from that AST.  Don't do that for this, but might be good to discuss.

fhackett:
@mjp41 all done I think

PR #127, opened 2024-07-19
Let Location::str() handle `\r` correctly

fhackett:
This PR is split off from #126, because really it's an orthogonal fix.

The observed problem was that, on Windows, error messages had a strange `\r` in reported source locations that was inconsistent with both other platforms, and other parts of the same output string (you would see `\n` once then `\r\n` further along the same string).

The trouble was traced back to `source.h`, which was separating lines based on only the `\n` character, and assuming that line breaks were only 1 char wide. This PR rewrites the 3 line/columns handling functions that did this, so that they will accept `\r\n`, `\n`, and `\r` as line delimiters while reporting correct line lengths (and so forth) on all platforms.

The main change of substance is that the `lines` field of `SourceDef` changes from a vector of `size_t` to a vector of `pair<size_t, size_t>`, because it is no longer possible to derive where a line ends from where another line begins (we might have to add/subtract either 1 or 2 chars). Rather than try to detect some file-wide line break width (and possibly have problems due to corrupted files mixing line ending types), this rewrite just looks at which line ending variant each line had and records span information to match.

Additionally, this PR adds exhaustive generative testing of this changed feature.
The test code is commented to explain logically how it builds all possible permutations of line endings and line contents up to a certain length (albeit using just `a` as contents since this code is not sensitive to string contents).
As configured in the `CMakeLists.txt`, it runs 729 test cases up to length 6 (6 chars, 6 line endings, and everything in between).

This fine-grained testing should help keep this code robust if it ever needs to change again - in fact it already helped catch errors when implementing some review recommendations from the other PR this was split from.

### DIFF
--- a/include/trieste/source.h
+++ b/include/trieste/source.h
@@ -81,8 +81,11 @@ namespace trieste
       // Find the first line that begins _after_ pos, and then backtrack one
       // element if we need to. We can't write this directly because "last
       // element for which condition was false" is not an stdlib primitive.
-      auto it = std::find_if_not(
-        lines.begin(), lines.end(), [&](std::pair<size_t, size_t> elem) {
+      auto it = std::lower_bound(
+        lines.begin(),
+        lines.end(),
+        pos,
+        [&](std::pair<size_t, size_t> elem, size_t pos) {
           return elem.first <= pos;
         });
       // If we're at the beginning already, or we couldn't find a line after
@@ -94,7 +97,7 @@ namespace trieste
         --it;
       }
 
-      size_t line = it - lines.begin();
+      size_t line = std::distance(lines.begin(), it);
       size_t col = pos - it->first;
 
       return {line, col};
--- a/include/trieste/source.h
+++ b/include/trieste/source.h
@@ -210,8 +210,9 @@ namespace trieste
             source->view().substr(interim_pos, linepos2 - interim_pos);
         }
 
-        write_indexed_skipping_r(
-          line_view_first, [&](size_t idx) { return idx < col ? ' ' : '~'; });
+        write_indexed_skipping_r(line_view_first, [&ccol = col](size_t idx) {
+          return idx < ccol ? ' ' : '~';
+        });
         ss << std::endl;
         write_chars_skipping_r(line_view_first);
         write_chars_skipping_r(interim_view);
@@ -230,7 +231,7 @@ namespace trieste
         assert(pos >= linepos);
         write_indexed_skipping_r(
           line_view.substr(0, pos - linepos + len),
-          [&](size_t idx) { return idx < col ? ' ' : '~'; });
+          [&ccol = col](size_t idx) { return idx < ccol ? ' ' : '~'; });
         ss << std::endl;
       }
 
--- a/include/trieste/source.h
+++ b/include/trieste/source.h
@@ -136,6 +136,9 @@ namespace trieste
       // print exactly the line and not fragments of \r\n.
       auto try_match = [](std::string_view curr, std::string_view part)
         -> std::optional<std::string_view> {
+        // substr will always return a valid result if the index is in range.
+        // Even if part.size() is out of bounds, we will just get a truncated
+        // string_view to compare against.
         if (curr.substr(0, part.size()) == part)
         {
           return curr.substr(part.size());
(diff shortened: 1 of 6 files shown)


include/trieste/source.h
```diff
@@ -71,46 +71,100 @@ namespace trieste
...
-      return {start, end - start};
+      return lines[line]; // already in {start, size} format
     }
 
   private:
     void find_lines()
     {
-      // Find the lines.
-      auto pos = contents.find('\n');
+      using namespace std::string_view_literals;
```
matajoh, line 132: As mentioned on the other PR (repeating comment here) I think a slight modification makes this a bit cleaner:

```c++
auto is_eol = [](const std::string_view s) -> std::pair<bool, size_t> {
        if(s.size() > 1 && s.substr(0, 2) == "\r\n"){
            return {true, 2};
        }
        
        if(s.empty()){
            return {false, 0};
        }
       
        if(s[0] == '\n'){
            return {true, 1};
        }
        
        return {false, 1};
    };

    size_t cursor = 0;
    size_t line_start = 0;
    std::string_view contents_view(contents);
    while(cursor < contents_view.size()){
        auto [eol, inc] = is_eol(contents_view.substr(cursor));
        if(eol){
            lines.emplace_back(line_start, cursor);
            cursor += inc;
            line_start = cursor;
        }else{
            cursor += inc;
        }
    }

    if(cursor - line_start > 0){
        lines.push_back({line_start, cursor});
    }
```

include/trieste/source.h
```diff
@@ -71,46 +71,100 @@ namespace trieste
...
+        {
+          return false;
+        }
+      };
+
+      size_t line_start = 0;
+      while (cursor < contents.size())
+      {
+        size_t last_pos = cursor;
+        if (try_match("\r\n"sv) || try_match("\n"sv) || try_match("\r"sv))
```
matajoh, line 153: Just a quibble but a line cannot end with carriage return. The only valid line endings are `\r\n` or `\n`. As unlikely as it may be, we do not want to be line breaking on `\r`.

matajoh (changes requested):

include/trieste/source.h
fhackett: I saw someone on StackOverflow talking about an old Mac system that had `/r` as a line break... which upon further inspection was OS 9 and has been obsolete since 2001.

Fair enough, my bad for not checking that more thoroughly. Easy enough to remove.

include/trieste/source.h
fhackett: Style question, now I realize it was the `[&]` you had a problem with. Why is that considered bad?

To me, a big strength of local functions is that you don't have to waste time with convoluted state passing when you can just reference local state (the local function has the same or shorter lifetime than the state, so no memory problems).  I use this a lot in Scala.

If there's a cultural difference in C++, I'd like to understand it properly, because almost every non-trivial accumulator-using function I've written while working on this project uses this style. If I knew what was seen as undesirable about it, I could make a broader change.

include/trieste/source.h
fhackett: Removed.

include/trieste/source.h
fhackett: While I would still like your thoughts on the above, I think the idea to use a string_view as an accumulator is good, and I updated my code.

I simplified it a bit from your example, since `substr` is quite robust, and I changed `try_match` to return the already-shifted string_view so that the position increment and the comparison are just next to each other, not inside and outside the function (easier to see that one guards the other). I used an optional ensure the null case, where the string didn't match, also works.

fhackett: (added a code comment for those who haven't just read the substr edge case docs)

fhackett: This has become mostly irrelevant, given subsequent discussion.

fhackett:
Based on meeting feedback, this new version hardens `Location::str()` against carriage returns instead.

Because I was exhaustively testing and found a lot of edge cases, I rewrote most of Location::str() in order more fundamentally satisfy the testing requirements.
That said, I am relatively sure that, if you don't use `\r`, the old and new implementation are exhaustively equivalent in their behavior. It can be double checked by reverting my `source.h` changes and running the tester with `--skip-carriage-returns`.

Let me know if there are any questions/clarifications/requests.

matajoh (approved):

PR #128, opened 2024-07-24
Use intrusive reference counting

fhackett:
This proposed change relates to the ongoing discussion of #115, regarding how Trieste can reduce its impact on binary size.

One thing we identified was the footprint of shared_ptr destructors, which will be included whenever the ubiquitous `Node` goes out of scope. We considered switching to a customized intrusive reference counted pointer implementation, which might be a bit more efficient, and would give us more control over code generation of such a common case in code that uses Trieste.

The resulting pointer implementation, and its non-inline refcount decrement, saves 300KB on the `yamlc` executable, and saved 1.3MB on rego-cpp's main executable, or about 20% of its original size (6.8MB). Keeping the refcount decrement inlineable (but otherwise the same) saves some space still, but less (200KB for `yamlc`).

Notes:
- ~~there is no support for multithreading in this changeset so far. We don't really need it for now, but perhaps best to add it as a compile-time configuration option.~~ not any more
- ~~question: the thread-safe intrusive ptr is disabled right now; what kind of option would make sense for enabling it?~~ answer after discussion: always enable thread-safety. Implemented.
- ~~this is a breaking change wherever a client was aware of aliases like `Node` being `shared_ptr` specifically. One notable example is the `shared_from_this()` method, which has an equivalent in the new version but is renamed. We could simplify the change by keeping the "shared" name, or we could go with a generic name like `ptr_from_this` to be robust against this problem in the future.~~ answer after discussion: current name is ok, change behavior of `->parent()` in other PR

### DIFF
--- a/include/trieste/intrusive_ptr.h
+++ b/include/trieste/intrusive_ptr.h
@@ -37,7 +37,7 @@ namespace trieste
         return value;
       }
 
-      copyable_refcount operator+=(size_t inc)
+      copyable_refcount& operator+=(size_t inc)
       {
         value += inc;
         return *this;
--- a/include/trieste/intrusive_ptr.h
+++ b/include/trieste/intrusive_ptr.h
@@ -26,14 +26,11 @@ namespace trieste
     SNMALLOC_SLOW_PATH
     constexpr void intrusive_dec_ref()
     {
-      if (intrusive_refcount == 1)
+      assert(intrusive_refcount > 0);
+      intrusive_refcount -= 1;
+      if (intrusive_refcount == 0)
       {
-        intrusive_refcount = 0;
         delete this;
-      }
-      else
-      {
-        intrusive_refcount -= 1;
       }
     }
 
@@ -111,8 +108,8 @@ namespace trieste
 
     constexpr intrusive_ptr<T>& operator=(const intrusive_ptr<T>& other)
     {
-      // Sets us to nullptr and holds onto our ptr in tmp
-      intrusive_ptr<T> tmp = std::move(*this);
+      // Hold onto our ptr until we return, adding 1 to refcount
+      intrusive_ptr<T> tmp{*this};
       ptr = other.ptr;
       inc_ref();
       // dec_ref for our original ptr goes here, where tmp gets destroyed.
@@ -125,7 +122,6 @@ namespace trieste
     constexpr intrusive_ptr<T>& operator=(intrusive_ptr<T>&& other)
     {
       std::swap(ptr, other.ptr);
-      other.dec_ref();
       return *this;
     }
 
--- a/include/trieste/intrusive_ptr.h
+++ b/include/trieste/intrusive_ptr.h
@@ -11,11 +11,44 @@
 
 namespace trieste
 {
-  enum class intrusive_ptr_threading
+  namespace detail
   {
-    sync,
-    async,
-  };
+    // In principle, std::atomic should not be copied.
+    // It should be a single object that is pointer-to and manipulated by
+    // multiple threads. For refcounts however, it should be possible to copy a
+    // refcounted object. The catch is that _everything but the refcount should
+    // be copied_. The copy constructors here will just set the new refcount to
+    // 0, as if the object was constructed from scratch, so different
+    // intrusive_ptr can take ownership of the new object.
+    struct copyable_refcount final
+    {
+    private:
+      std::atomic<size_t> value = 0;
+
+    public:
+      constexpr copyable_refcount(size_t value_) : value{value_} {}
+
+      constexpr copyable_refcount() = default;
+      constexpr copyable_refcount(const copyable_refcount&) : value{0} {}
+      constexpr copyable_refcount(copyable_refcount&&) : value{0} {}
+
+      constexpr operator size_t() const
+      {
+        return value;
+      }
+
+      constexpr copyable_refcount operator+=(size_t inc)
+      {
+        value += inc;
+        return *this;
+      }
+
+      constexpr size_t fetch_sub(size_t dec)
+      {
+        return value.fetch_sub(dec);
+      }
+    };
+  }
 
   template<typename T>
   struct intrusive_refcounted_traits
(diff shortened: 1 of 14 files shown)


include/trieste/intrusive_ptr.h
```diff
@@ -0,0 +1,341 @@
...
+      // Self-assignment case (skip the code below that would destroy the object
+      // otherwise)
+      if (ptr == other.ptr)
+      {
+        return *this;
+      }
+
+      dec_ref();
+      ptr = other.ptr;
+      inc_ref();
```
mjp41, line 128: I think this code might be wrong.  The `other` is effectively borrowed as you are using a reference.  So I think it is possible for the call to decref to run a destructor, which in turn deallocates the underlying object that `other` is borrowed from.

You really should do
```suggestion
      T* tmp = ptr;
      other.ptr.intrusive_inc_ref();
      ptr = other.ptr;
      tmp.intrusive_dec_ref();
```

This should cover the self-assignment case.  You might need null-checks too.

include/trieste/parse.h
```diff
@@ -533,13 +534,13 @@ namespace trieste
   inline detail::Rule
   operator>>(detail::Located<const std::string&> s, detail::ParseEffect effect)
   {
-    return std::make_shared<detail::RuleDef>(s, effect);
+    return intrusive_ptr(new detail::RuleDef(s, effect));
```
mjp41, line 537: I don't like this pattern.  Raw `new` is generally considered bad form.  Could we wrap is and expose something like:
```C++
template<typename T, typename... Args>
intrusive_ptr<T> intrusive_refcounted<T>::make(Args&&... args) {
   return intrusive_ptr(new T(std::forward<Args>(args)...);
}
```

Or a `make_intrusive` function if the template and inheritance don't work out with it being a method.

include/trieste/intrusive_ptr.h
fhackett: ... yes, I think I see it.

All you need is (in some weird syntax that mixes up C++, Scala, and probably Go, sorry):
```
let foo = X
let bar = Cons(foo, Nil)

foo := Nil // bar indirectly has only ref to what foo had
bar := bar.head // destroys Cons(...) -> destroys X -> bar now points to bad
```

fhackett: Actually my previous version was basically right then, in that what you're suggesting here is equivalent up to some syntax.
The complaint was that it was doing a redundant inc_ref on self-assign, and I introduced this bug when removing the self-assign handling.

So, I put my existing solution back but with the self-assign "optimization" still there, since it also gets all the null checks right by construction (rather than give me more pointer manipulation to shoot myself in the foot with).

(working on the other comment, then committing)

include/trieste/parse.h
fhackett: I got intrusive_ptr<T>::make working.

I could use it _almost_ everywhere, except where it was never possible due to NodeDef's private constructors.

include/trieste/intrusive_ptr.h
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    friend std::hash<intrusive_ptr<T>>;
+  };
+
+  template<typename T>
+  struct intrusive_refcounted
+  {
+  private:
+    detail::copyable_refcount intrusive_refcount = 0;
```
mjp41, line 196: The assignment can end up calling a copy constructor, it is better to use the initialisation syntax `{foo}` rather than `= foo`.
```suggestion
    detail::copyable_refcount intrusive_refcount{1};
```
Also, a more common pattern is to start with `1`, and then you don't need an incref on construction.  The incref should could get optimised, but if it is using atomics, then it is probably not allowed to optimise.

include/trieste/intrusive_ptr.h
```diff
@@ -0,0 +1,350 @@
...
+      {
+        return *this;
+      }
+
+      intrusive_ptr<T> tmp;
+      // Don't actually inc_ref, but putting old ptr in tmp lets us leverage the
+      // built in dec_ref with null checks below.
+      tmp.ptr = ptr;
+      ptr = other.ptr;
+      inc_ref();
```
mjp41, line 136: It is more common to put the `inc_ref` above the assignment.  The implementation does not handle the read-reclaim race, so is not too important.  How about?
```C++
other.incref();
// Move old value into a temporary, so gets dec_ref after assignment.
intrusive_ptr<T> old{ptr};
ptr = other.ptr;
```

Note this assumes initial `rc = 1` change proposed belove

FYI, I initially didn't like the use of the destructor, but I am coming around to it.

include/trieste/intrusive_ptr.h
```diff
@@ -0,0 +1,350 @@
...
+      return intrusive_ptr(new T(std::forward<Args>(args)...));
+    }
+
+    constexpr intrusive_ptr() : ptr{nullptr} {}
+
+    constexpr intrusive_ptr(std::nullptr_t) : ptr{nullptr} {}
+
+    constexpr explicit intrusive_ptr(T* ptr_) : ptr{ptr_}
+    {
+      inc_ref();
```
mjp41, line 103:
```suggestion
```
If we start with `rc=1`, then this can be removed.

include/trieste/intrusive_ptr.h
```diff
@@ -0,0 +1,350 @@
...
+      tmp.ptr = ptr;
+      ptr = other.ptr;
+      inc_ref();
+      // tmp gets dec_ref here, potentially destroying the value at old ptr
+      return *this;
+    }
+
+    constexpr intrusive_ptr<T>& operator=(intrusive_ptr<T>&& other)
+    {
+      std::swap(ptr, other.ptr);
```
mjp41, line 143: Is this really a swap? Should we be putting `ptr` into `other`?

```suggestion
intrusive_ptr<T> old{ptr};  
ptr = other.ptr;
other.ptr = nullptr;
```

mjp41 (changes requested):
A few more minor bits.

include/trieste/source.h
```diff
@@ -115,6 +127,8 @@ namespace trieste
     }
   };
 
+  using Source = intrusive_ptr<SourceDef>;
```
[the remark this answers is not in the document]
fhackett: Not sure why that's there. Deleting.

include/trieste/source.h
```diff
@@ -2,26 +2,38 @@
...
+  // be found under NodeDef's definition.
+  template<>
+  struct intrusive_refcounted_traits<NodeDef>
+  {
+    static constexpr void intrusive_inc_ref(NodeDef*);
+    static constexpr void intrusive_dec_ref(NodeDef*);
+  };
+
+  using Source = intrusive_ptr<SourceDef>;
+  using Node = intrusive_ptr<NodeDef>;
```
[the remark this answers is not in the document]
fhackett: A good question that I largely avoided for this PR up to this point (didn't want more compile errors because I broke something that wasn't the ptr implementation).

It can be moved one file "down" to token.h, but it stops there because there is a circular relationship between TokenDef and NodeDef.

That is probably cleaner, and it does work. I'll do that.

include/trieste/intrusive_ptr.h
fhackett: Ok, so my existing design comes from the idea that we should support "ingesting" a raw pointer into an intrusive_ptr, whether it's owned or not.

If we start at refcount 1 instead, then we have to know if we're on the first construction (in which case no increment), or if we're taking joint ownership (in which case, do increment) (yes via a raw pointer, yes rare).

This is not a problem for other refcounts where we control the initialization more. For example, shared_ptr completely hides its control block's lifecycle, so it can probably just start at 1 with no problem. Our control block is just the object itself, and we let users allocate one whenever they want.

The trouble is, if we ever write `intrusive_ptr{new X}`, and X's constructor started its refcount at 1, then we might end up getting a refcount of 2. Conversely, if we assume the raw ptr construction must have a refcount of 1 (though that is assertable I suppose), then it's possible to mess up and not increment in some cases (which admittedly are usually in utility methods like `intrusive_ptr_from_this` or intrusive_ptr casting, but it is complexity nonetheless).

For context, there are unavoidable `intrusive_ptr{new NodeDef}` instances that are needed because the NodeDef constructor is private and `make` can't call it.

So basically I can implement start-at-1, but it will have some code complexity cost. What do you think?

include/trieste/intrusive_ptr.h
fhackett: (leaving out the rc init consideration which I wrote about a bit above)

Interesting, hadn't thought about that dimension. (read-reclaim race... I can imagine what it means but not familiar)

I can do that, just need to sprinkle an extra `const` on `inc_ref` (which I think is fine because const semantics).

include/trieste/intrusive_ptr.h
fhackett: I've seen this done as a swap somewhere, but nulling other is fine too. Just trying to be "fancy" and I'm not too attached to it.

I can do it the other way.

include/trieste/intrusive_ptr.h
fhackett: Resolved via voice call:
- keep starting value at 0
- comment why, summarizing this conversation
- make sure the 0 only shows up once (likely, remove copyable_refcount's default constructor)

include/trieste/intrusive_ptr.h
mjp41: `std::exchange` would probably be the correct "fancy".  Please don't try to be fancy.

fhackett: It was a specific thing I remember from an undergrad C++ course actually, something about re-using constructor/destructor to implement assignment. Yes, specifically with `swap`... but I either misremembered it, or it's not useful here.

Not so much "trying" to be fancy, as ending up doing it without realizing occasionally.

mjp41 (approved):

PR #131, opened 2024-08-09
Fix intrusive_ptr test breakage in dependencies

fhackett:
Does what the title says.

I noticed when trying to update rego-cpp that the tests, when auto-run from a dependency, fail because they can't find their executables.

This didn't show up before because Trieste did not itself have tests until I added this last set (all tests were indirect and optional, part of the tools).

It might actually be more reasonable to disable running tests when included as a dependency, but I'm not sure how to do that. Looking around on the internet, with my search terms at least, is inconclusive.

This change on its own at least makes sure the tests can be launched correctly regardless of context.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -21,6 +21,7 @@ jobs:
         include:
         - platform: "ubuntu-latest"
           variant: "ninja"
+          build-type: "Release"
           compiler: "clang"
           cmake-options: "-DCMAKE_CXX_COMPILER=clang++-15 -DCMAKE_C_COMPILER=clang-15 -DTRIESTE_BUILD_PARSER_TESTS=1"
           generator: "-G Ninja"
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ "main" ]
 
+env:
+  global-cmake-flags: -DTRIESTE_ENABLE_TESTING=1
+
 jobs:
   build-test:
     strategy:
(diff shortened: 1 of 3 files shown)


fhackett:
(there seems to be no problem with the build, just another stackoverflow from the fuzzer exploring beyond usable stack depth)

fhackett:
Mysteriously, adding `-DTRIESTE_ENABLE_TESTING=1` to the `ubuntu-latest clang` (but notice the missing "Release"/"Debug" annotation) CI run causes a breakage because `build-type` is missing.

I'm not sure how that ever worked, because it is always assumed that there is a build type, and that entry didn't have one. Leaving it unspecified would mean "Debug", and the unique feature of that CI entry is to use Ninja, so I specified the build type and marked the variant name as "ninja".

fhackett:
Some more experiments later, conclusion:
- put `-DTRIESTE_ENABLE_TESTING=1` unconditionally in the config command itself. Putting it in the matrix means that overrides will not pick it up without copy-pasting it _sometimes_. For something truly unconditional, I think best just write it in the command.
- Not writing some field names in the matrix is not only not a mistake, it's semantically meaningful. The field not being in the matrix def means that includes that specify that field will mutate existing matrix entries where possible, as opposed to adding new ones. Adding even an empty string entry for such a field will cause additional malformed CI jobs to be generated. I commented this realization in the YAML.

.github/workflows/buildtest.yml
```diff
@@ -62,10 +64,10 @@ jobs:
     - uses: actions/checkout@v3
     
     - name: Install build dependencies
-      run: ${{ matrix.dependencies }}
+      run: ${{matrix.dependencies}}
       
     - name: Configure CMake
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{matrix.build-type}} ${{matrix.generator}} ${{matrix.standard}} ${{matrix.cmake-options}}
+      run: cmake -B ${{github.workspace}}/build -DTRIESTE_ENABLE_TESTING=1 -DCMAKE_BUILD_TYPE=${{matrix.build-type}} ${{matrix.generator}} ${{matrix.standard}} ${{matrix.cmake-options}}
```
matajoh, line 70: I think you could add "TRIESTE_BUILD_PARSER_TESTS=1" here as well. It might be cleaner to have `${{matrix.testing}}` here and then add that in the matrix above to make it clearer/easier to add testing-specific flags in the future.

fhackett: If it's ok to remove the current matrix of having it both enabled and disabled in different CI runs, sure.

Re: matrix below

fhackett:
> I'm quite keen on this `matrix.testing` change if you're up for it.

Given my previous findings, I'm not sure it makes sense. Putting that config in "matrix" strictly means it won't always apply. Any CI run added by `includes`, not changed by `includes` (of which there are a few), will not use the field. If there was way to make just a non-matrix variable I'd be all for this, but if it has to be in "matrix" then it doesn't work as intended.

fhackett:
Well, turns out there was a way to make a constant variable outside of the matrix: `env: x: ...`, which is then accessed as `${{env.x}}`.

matajoh (approved):

PR #132, opened 2024-08-11
Generalized recursion-safe intrusive_ptr destructor

fhackett:
As discussed when #128 was reviewed, there was a question about generalizing the recursion-safe destructor logic in `trieste::NodeDef` to the pointer type itself.

This PR implements that proposed change, while dodging an unforeseen pitfall of generalizing those semantics.

Namely, not all `intrusive_refcounted<T>` live on the heap. Moving the `NodeDef` destructor which (correctly) assumes all its instances are heap-allocated, to classes that are often but not always heap allocated, ends up trying to free stack-allocated memory. The work-around is to move the destructor logic somewhere that only runs when `T*` is guaranteed to be heap allocated, which is `intrusive_dec_ref` -- because once you give a `T*` to `intrusive_ptr<T>` then you are claiming that heap memory management is ok.

I added a private helper with all of the destructor logic called `recursion_safe_delete_this`, which is called instead of the raw `delete this` in `intrusive_dec_ref`. The strategy is otherwise unchanged.

I also made an effort to update the documentation comment so that it a) reflects the generalized context, and b) explains why the code needs to be called from `intrusive_dec_ref` and not `~intrusive_refcounted<T>`.


fhackett:
(and, as CI reminded me, it requires removing `constexpr` in a couple of places... although the entire setup would be valid constexpr with C++23, apparently; the issue is the `thread_local` qualifier)

sylvanc (approved):

PR #135, opened 2024-08-21
Add tooling to debug pattern matching.

xFrednet:
It can be hard to determine why a certain pattern isn't matching with the given tree. This PR adds a `Dbg` function which can be wrapped around the pattern and effect to debug the matching step.

The function has a custom pattern that wraps around the original pattern. It then counts how many nodes have been matched until the match failed. The debug macro has a defined `start` token, that *engages* the debugging. The results are remitted as error nodes to the AST.

<details>
<summary>Example</summary>

Let's consider the following AST as an example:

```
(if)
(int 1:5)
(leq)
(int 1:6)
(then)
(int 2:10)
(else)
(int 3:100)
``` 

I wrote the following rule, to match the expressions:

```c++
T(If)
    * (((Any * --T(Then))++ * Any)[Op]) * T(Then)
    * (((Any * --T(Else))++)[Then]) * T(Else)
    * Any++[Else] >>
  [](Match& _)
  {
    // [..]
  },
```

However, the match sadly fails. Now I can add the `Dbg` function as a wrapper:

```diff
+ Dbg(If,
 T(If)
     * (((Any * --T(Then))++ * Any)[Op]) * T(Then)
     * (((Any * --T(Else))++)[Then]) * T(Else)
     * Any++[Else] >>
   [](Match& _)
   {
     // [..]
   },
+ ),
```

The first `If` indicates that the debugging should start at the first `If` token it encounters. The function generates the following output:

```
(error
    (debugmsg 26:match failed after 5 nodes)
    (debugmsg 20:capture 'then': )
    (debugmsg 21:capture 'op': 
        (int 1:5)
        (leq)
        (int 1:6))
    (debugmsg 13:matched nodes
        (if)
        (int 1:5)
        (leq)
        (int 1:6)
        (then)))
    (int 2:10)
    (else)
    (int 3:100))
```

It's indicated that the first five nodes `(if)(int)(leq)(int)(then)` could be matched.

</details>

### Optional Improvements:

* The console output currently only prints the message. Is there a simple way to show the children nodes and also show the "source code" representation of what has been matched?
* Document the function in a more visible place

---

My C++ is a bit rusty, so there are probably several unidiomatic expressions. I'll be happy to fix them.

I also encountered a problem with `intrusive_ptr` where I couldn't pass `intrusive_ptr<DebugPattern>` into a function expecting `PatternPtr`. Is there a simple way to do this conversion?

Also, I wasn't sure how to properly test this, since the usage of this function should be an exception and not the rule. :sweat_smile: 

---

cc: @EliasC


xFrednet:

include/trieste/rewrite.h
```diff
@@ -1191,4 +1303,77 @@ namespace trieste
...
+    detail::PatternPtr pattern = intrusive_ptr<detail::DebugMatch>::make(
+      type, effect.first.value, dbg_on_success);
+    auto original_effect = effect.second;
+    return {
+      detail::Pattern(
+        pattern,
+        // We want to accept any token, since the contained pattern could
+        // start with any token, and when want to be invisible until the
+        // start token was found
+        detail::FastPattern::match_pred()),
```
mjp41, line 1348: I think this could be 
```suggestion
        detail::FastPattern::match_token({type})),
```

Not sure it is worth changing as this is a Debug feature so perf is probably not an issue.

include/trieste/rewrite.h
```diff
@@ -1035,6 +1041,112 @@ namespace trieste
...
+      : start(start_), pattern(pattern_), dbg_on_success(dbg_on_success_)
+      {}
+
+      PatternPtr clone() const& override
+      {
+        throw std::runtime_error("The DebugMatch pattern can't be cloned");
+      };
+
+      bool match(NodeIt& it, const Node& parent, Match& match) const& override
+      {
```
mjp41, line 1072: I'm unsure on what this will produce in the failure case for a disjunction?  Will it only report the failure of the final disjunct?

mjp41 (commented):
@xFrednet thanks for looking into this.  I am unclear on the precise aims of this.  As you are only generating pretty print on the final state of a match can we get the same information in the debugger without code changes?

Adding logging to the pattern matching code be very helpful for investigating why something failed internally.

Does this provide the part of the pattern that matched, and the part that didn't?

I have been wondering if we should have an AST for rewrite patterns, and I think that could help with this kind of debugging experience as you would be able to return the continuation that has not matched.  Currently there is no printable representation for that.

include/trieste/rewrite.h
xFrednet: Yes, this is sadly a limitation of this non-intrusive approach.

include/trieste/rewrite.h
xFrednet: If the wrapped pattern can start with `T(type1, type2)` wouldn't `detail::FastPattern::match_token({type1}))` prevent the pattern being matched on `type2`?

I basically tried to make it still work for every token besides the one set in the `Dbg` function

mjp41: Ah, yes you are right

PR #142, opened 2024-09-23
Support for Reapply effect

### LINKED ISSUE #139: A case for immediately matching on inserted nodes
If you have `A B C` and apply the rules `B ~> D` and `D ~> E` in a `once` pass, you will end up with `A D C` since matching continues *after* the newly inserted `D` node. This is often what you want, especially in a once pass since it lets you modify nodes uniformly without getting stuck in a loop. However, there are cases where it makes sense to produce a value that can immediately be handled in the same pass. This came up in @matteB10's work on constraint solving for MiniML, where for example a constraint `A -> B = C -> D` reduces to two constraints `A = C` and `B = D`. If they could be inserted "to the right" of the thing being replaced instead of "to the left", they would be processed in the same iteration, allowing the pass to be a `once` pass. 

Maybe it is possible to get a handle to the tail of the current matching position and insert nodes manually there, but it would be nice to be able to do this within Trieste itself, for example by having an effect like `Continue << node` to insert `node` so that it is the next thing being matched on.

EliasC:
This PR adds an effect `Reapply` that makes matching continue at the newly inserted node(s). It is intended to be used in `once` passes for "local fixpoint" computations but can be used with non-`once` passes as well. `Reapply` works like `Seq` in that the node is replaced by all of its children. As an example, consider the rewrite rules

    T(A) * T(A) >> [](auto&) -> Node { return B; }
    T(B) * T(A) >> [](auto&) -> Node { return C; }

When run in a `once` pass they would turn "**A A** A" into "B A". Changing the effect in the first rule to `return Reapply << B` would make the rewrites in sequence be "**A A** A ~> **B A** ~> C". Of course, careless use of `Reapply` can make a `once` pass loop forever.

Closes #139.

### DIFF
--- a/include/trieste/token.h
+++ b/include/trieste/token.h
@@ -174,19 +174,19 @@ namespace trieste
   inline const auto Group = TokenDef("group");
 
   // Special tokens for effects
-  inline const auto Seq = TokenDef("seq");
+  inline const auto Seq = TokenDef("seq", flag::internal);
   inline const auto Lift = TokenDef("lift", flag::internal);
-  inline const auto NoChange = TokenDef("nochange");
+  inline const auto NoChange = TokenDef("nochange", flag::internal);
   inline const auto Reapply = TokenDef("reapply", flag::internal);
 
   // Special tokens for symbol tables
-  inline const auto Include = TokenDef("include");
+  inline const auto Include = TokenDef("include", flag::internal);
 
   // Special tokens for error handling
-  inline const auto Invalid = TokenDef("invalid");
+  inline const auto Invalid = TokenDef("invalid", flag::internal);
   inline const auto Error = TokenDef("error", flag::internal);
-  inline const auto ErrorMsg = TokenDef("errormsg", flag::print);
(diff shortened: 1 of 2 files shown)


include/trieste/token.h
```diff
@@ -167,15 +167,23 @@ namespace trieste
     constexpr TokenDef::flag internal = 1 << 6;
   }
 
-  inline const auto Invalid = TokenDef("invalid");
+  // Built-in grouping
```
EliasC, line 170: While boyscouting here: Is there a reason for `Seq` and `NoChange` to not be `internal`?

mjp41: I think they should be.  When I added internal it was for a particular bit of code, but it makes sense to extend that here.

mjp41 (approved):
LGTM

sylvanc (approved):

EliasC:
It seems like the GNU compilers don't like comparing a typed enum to a non-enum... I will revert to using `const variables` to make CI happy

EliasC:
@mjp41 CI is happy again!

PR #146, opened 2024-09-26
Add support for removing shapes from WF specifications

EliasC:
This PR adds support for removing shapes from a WF specification:

```
inline const auto wf1 =
  (Foo <<= Bar++)
| Baz <<= Foo * Bar
;

inline const auto wf2 =
  (wf1 - Foo)
  | Bar <<= Baz++
```

In `wf2` `Foo`s can still appear in the tree, but they must have zero children (just like if they were never in the WF to begin with). Without this PR, the only way to achieve the same functionality is by repeating the shape list without `Foo`.

To be honest, I'm not sure what I'm doing with the functions taking rvalues, I'm just parroting other similar functions.

### DIFF
--- a/include/trieste/wf.h
+++ b/include/trieste/wf.h
@@ -1071,6 +1071,30 @@ namespace trieste
         wf.append(shape2);
         return wf;
       }
+
+      inline Wellformed operator-(const Wellformed& wf, const Token& token) {
+        Wellformed wf2;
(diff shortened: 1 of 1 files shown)


mjp41 (approved):

mjp41:
LGTM

sylvanc (approved):

PR #149, opened 2024-11-26
Added upper bounds for sequences

EliasC:
This PR adds support for giving upper bounds for sequences in the well-formedness specifications:
```
inline const auto some_wf =
  (File <<= Group++) // Zero or more groups
| (Foo <<= Group++[1]) // One or more groups
| (Bar <<= Group++[1][3] // Between one and three groups
```

One motivating use case is having a top-level optional shape (which is shorthand for `++[0][1]`):
```
inline const auto foo_wf =
  some_wf
  | (Paren <<= ~(Foo | Bar)) // Optionally contains a single Foo or a single Bar
```
Before this, the closest you could get was using `++`, which would mess with fuzz testing. Just as sequences, optional shapes can only be used at the top level of a shape; you still cannot write e.g. `Paren <<= Foo * ~Bar` or `Paren <<= Foo * Bar++`.

The JSON parser had `File <<= Group++` even though the initial parse pass can produce at most one group, but I did not find any more interesting usages of this pattern in any other sample or parser. Manual toying around with this seems to suggest that both checking and generation handles the new shape correctly.

### DIFF
--- a/include/trieste/wf.h
+++ b/include/trieste/wf.h
@@ -247,9 +247,11 @@ namespace trieste
     {
       Choice choice;
       size_t minlen;
+      size_t maxlen;
+      bool has_minlen = false;
 
-      SNMALLOC_SLOW_PATH Sequence(Choice choice_, size_t minlen_) :
-        choice{choice_}, minlen{minlen_} {}
+      SNMALLOC_SLOW_PATH Sequence(Choice choice_, size_t minlen_, size_t maxlen_) :
+        choice{choice_}, minlen{minlen_}, maxlen(maxlen_) {}
       SNMALLOC_SLOW_PATH Sequence() = default;
       SNMALLOC_SLOW_PATH Sequence(const Sequence&) = default;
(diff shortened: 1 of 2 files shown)


EliasC:
I have updated this PR to generalise the `Sequence` shape instead of special casing an `Option` shape.

include/trieste/wf.h
```diff
@@ -262,9 +264,17 @@ namespace trieste
...
+        if (!has_minlen)
+        {
+          minlen = new_len;
+          has_minlen = true;
+        }
+        else
+        {
+          maxlen = new_len;
+        }
         return *this;
```
mjp41, line 278: Should we detect
```C++
Foo++[1][2][3]
```
as an error?  Would we error for maxlen < minlen?

EliasC: Neither of the errors give a strange behavior, but I guess it would be a slightly nicer user experience to check for it. I will add it...

EliasC:
Added bounds checking (`min_len <= max_len` and no more than two bounds) now!

sylvanc (approved):

PR #153, opened 2025-02-19
Statistics and other general improvements of testing

EliasC:
This PR takes the first steps towards evaluating the generated trees during testing. The main addition is printing statistics about each pass, including how often each error rule was hit, how many times the pass succeeded, whether a successful pass actually touched the tree, and how many (hash) unique trees were generated. As a direct result of this statistics, we found (and fixed) that some starting seeds would result in the same minimal tree being generated tens of thousands of times in a row.

The concrete additions in this PR are as follows:
- Testing with the `Info` log level prints statistics as outlined above.
- The generator warms up the random number generator to get better entropy across adjacent seeds. For regression testing, the `test` command now supports the subcommand `debug_entropy` which will measure the per-byte entropy of a specified number of adjacent seeds after a specified amount of warmup.
- Testing will now store the hash of each generated tree and retry generation if there is a collision. The maximum number of retries is customizable.


EliasC:
This PR used to also add testing of passes in sequence. I removed that to make reviewing more manageable and will send a separate PR for that once this has gotten merged.

include/trieste/ast.h
```diff
@@ -762,6 +762,12 @@ namespace trieste
       return out.str();
     }
 
+    size_t hash()
+    {
+      // TODO: Implement something that doesn't go via a string
```
mjp41, line 767: Is this blocking for merge?

include/trieste/driver.h
```diff
@@ -109,6 +109,13 @@ namespace trieste
       bool test_failfast = false;
       test->add_flag("-f,--failfast", test_failfast, "Stop on first failure");
 
+      std::optional<size_t> test_max_retries = std::nullopt;
+      test->add_option("-r,--max_retries", test_max_retries,
+                       "Maximum number of retries for finding unique trees");
+
+      auto entropy = test->add_subcommand("entropy",
```
mjp41, line 116: This is not expected to be a user facing feature?  I wonder if we should call it `debug_entropy`.  Perhaps add to the description that this is intended for debugging the fuzzing phase coverage rather than for normal usage of a tool based on Trieste?

include/trieste/fuzzer.h
```diff
@@ -124,6 +126,87 @@ namespace trieste
...
+      double entropy = 0.0;
+      for (const auto& [byte, count] : freq) {
+        double p = count / total;
+        entropy -= p * std::log2(p);
+      }
+
+      return entropy;
+    }
+
+    int test_entropy() {
```
mjp41, line 159:
```suggestion
    int debug_entropy() {
```

mjp41 (approved):
LGTM

include/trieste/ast.h
EliasC: My feeling is that it is not blocking. Going via a string is slower, but I think we prefer slower fuzz testing on unique trees over fast fuzz testing on the same tree. I could remove the TODO and add it as an issue instead.

include/trieste/driver.h
EliasC: I think that is a good idea! Will fix

include/trieste/fuzzer.h
EliasC: Done in the latest commit

include/trieste/ast.h
EliasC: I removed the TODO and implemented a hash function that does not go via string :)

EliasC:
All comments have been addressed, @mjp41.

PR #154, opened 2025-03-03
Updated parser API with convenience in and blind pop

EliasC:
This PR adds two things to the parser API:

* Versions of `in` and `group_in` that takes a list of tokens, for checking if we are in one of several kinds of nodes.
* A version of `pop` that takes no argument, blindly popping from whatever node we are in and returning its token.

I find myself often wanting to implement "keep popping until we are no longer in one of these things" or "keep popping until we are inside one of these things". This makes that kind of logic simpler to implement (see diff of shrubbery parser for an example).

### DIFF
--- a/include/trieste/parse.h
+++ b/include/trieste/parse.h
@@ -84,12 +84,25 @@ namespace trieste
         return node == type;
       }
 
+      bool in(const std::initializer_list<Token>& types) const
+      {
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
LGTM.  Love how you managed to test it all in such a small change :-)

PR #156, opened 2025-04-16
Fixing tracking of changes in pass

hugolonnqvist:
While working with Trieste for my Bachelor's project, I noticed that sometimes the changes made during a pass do not seem to be registered. Upon looking into this, I found out that the number of changes tracked for a given match was determined by how many nodes were inserted in its place. The consequence of this, is that removal of nodes (i.e. matching but only returning {}) did not register as a change. 

This raises the question of what should be considered a change. The idea implemented in this PR, is that matching a pattern and modifying the AST in some way, (all matches which do not return NoChange) should be considered a single change. 

While fixing this I also added the total changes made by a group of passes as part of the return of the ProcessResult struct. This gives the ability to for example after a Rewriter has finished, inspect if any changes were made (and also how many). I felt this feature was kind of related to the fix and so it ended up in the same PR.

### DIFF
--- a/include/trieste/passes.h
+++ b/include/trieste/passes.h
@@ -118,7 +118,7 @@ namespace trieste
     std::string last_pass;
     Node ast;
     Nodes errors;
-    size_t total_changes;
+    size_t total_changes = 0;
 
     void print_errors(logging::Log& err) const
     {


hugolonnqvist:

EliasC (approved):
This looks good to me!

EliasC:
The build seems to crash due to the extra argument to ProcessResult not being supplied everywhere though.

mjp41 (approved):
LGTM, thanks for finding and fixing this.

PR #157, opened 2025-04-23
Check indexing when binding symbols

### LINKED ISSUE #140: Undescriptive runtime error when well-formedness with a symbol table entry fails
I have a well-formedness similar to this:

```c++
inline const trieste::wf::Wellformed flatten =
    (Top <<= File)
    | (File <<= (Label | Print)++)
    | (Label <<= Ident)[Ident];
    ;
```

Notice that the `Label` node has an identifier which should be added to the symbol table.

A rewrite rule returned something like this:

```c++
return Seq << (Print ^ " (True)")
           << (Label ^ "else:") << (Ident ^ else_label)
           << (Print ^ "(False"));
```

The label construction is incorrect, as I was missing parentheses to add the `Ident` as a `Child` of `Label`.

I expected the well-formedness to fail. Instead, I got the following undescriptive runtime error:

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
Aborted (core dumped)
```

Ideally, it should just throw the usual error about the well-formedness error

EliasC:
This PR fixes #140. You could get an out-of-bounds error when populating symbol tables if you generated a node with a binding (e.g. `(Assign <<= Ident * Expr)[Ident])`) with too few children. The solution here is to abort early and pretend that everything is fine. This allows checking for errors (maybe the binding was malformed due to the presence of errors) and relies on the WF check to reject the tree later. You cannot simply do the WF check first since that also checks that bindings have a corresponding entry in a symbol table.

### DIFF
--- a/include/trieste/wf.h
+++ b/include/trieste/wf.h
@@ -469,6 +469,11 @@ namespace trieste
         {
           if (field.name == binding)
           {
+            if (index >= node->size())
+              // Node does not have enough children. Pretend all is fine to
+              // allow error nodes, otherwise WF check will sort it out.
+              return true;
+
             auto name = node->at(index)->location();
 
             if (!node->bind(name))


mjp41 (approved):
LGTM

PR #158, opened 2025-04-24
Adding conditional passes

hugolonnqvist:
This PR introduces the ability to specify a `cond()` function on a pass. The goal is to allow a clear way to disable passes when needed. It's defined similar to the `pre()` and `post()` functions, with the difference that the lambda passed to the function, should return a boolean deciding if the pass should execute.

The `cond()` function can be defined inside a pass similar to the `pre()` and `post()` functions. It  may also be defined in a rewriter, which enables code shown below:

```cpp
Rewriter rewriter = {
  "optimization_analysis",
  {
      gather_instructions(control_flow),
      gather_flow_graph(control_flow),
      constant_folding(control_flow),

      gather_instructions(control_flow).cond(control_is_dirty),
      gather_flow_graph(control_flow).cond(control_is_dirty),
      dead_code_elimination(control_flow),
      dead_code_cleanup(),
  },
  whilelang::normalization_wf,
};
```
The context is that the `gather_instructions` and `gather_flow_graph` passes should not collect control flow information again unless its outdated.

### DIFF
--- a/include/trieste/pass.h
+++ b/include/trieste/pass.h
@@ -24,6 +24,7 @@ namespace trieste
   {
   public:
     using F = std::function<size_t(Node)>;
+    using CondF = std::function<bool(Node)>;
 
   private:
     static const int NOCHANGE = -1;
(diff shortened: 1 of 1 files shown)


mjp41 (approved):
LGTM.  I think working out how to fuzz individual passes with this would be very interesting.

Issue #125, opened 2024-06-26
Fuzzer failure: segfault in yaml_fuzz_to_json, seed 1597729744


fhackett:
This came up in an unrelated CI run.

From cursory investigation, the segfault happens deep inside snmalloc, while doing the seemingly harmless action of allocating a synthetic SourceDef, as of commit 917938ee73880db75efe738d39d5667566abd8f8.

I don't know what it means, and it will require either someone with more knowledge of snmalloc, or just a deeper investigation.

For convenience, here is my VSCode debug definition (your exact folder structure/args may vary, but know at least that this reproduces the problem on my machine):
```json
{
  "type": "lldb",
  "request": "launch",
  "name": "yaml_fuzzer to_json",
  "program": "${workspaceFolder}/out/build/debug-clang/parsers/test/yaml_fuzzer",
  "args": ["to_json", "-s", "1597729744", "-f"],
  "cwd": "${workspaceFolder}"
}
```

mjp41:
Can you run with Asan enabled. If you break inside snmalloc, then you probably have a memory safety bug.

fhackett:
I did that, and it seems this is a stack overflow in `Wellformed::gen_node`.

A little less scary than an allocator bug, at least. Thanks for the pointer.

I can investigate this later, unless someone gets to it first.

fhackett:
Another sighting in the #131  PR build. Seed was 4228657190

fhackett:
Another sighting in the #130 PR build. Seed was 1707913749

fhackett:
Another sighting in the #130 PR build. Seed was 2943010191.

PR #162, opened 2025-06-09
Do not export warning flags, but apply them to all targets the project builds

achamayou:
Instead of setting Warnings/Werror on the trieste interface target, which means that any code using the target picks up the flags, this PR is restricting it to the binaries built by the project (tests and sample parser libraries).

This is so that projects that use Trieste, directly and otherwise, as well as other code that is not necessarily warning-free, and that those other projects may not control, can build binaries effectively.

I have diffed the output of `ninja -j1 -v`, pre and post change, to make sure I did not miss anything. Note that projects depending on Trieste, such as https://github.com/microsoft/rego-cpp/ will need to do something similar, to keep building their tests/binaries with Warnings/Werror.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,12 +116,14 @@ if (TRIESTE_SANITIZE)
   target_link_libraries(trieste INTERFACE -fsanitize=${TRIESTE_SANITIZE})
 endif()
 
-if(MSVC)
-  target_compile_options(trieste INTERFACE /W4 /WX /wd5030 /bigobj)
-else()
-  target_compile_options(trieste INTERFACE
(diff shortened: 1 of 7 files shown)


achamayou:
Despite this change being believed to be a CMake no-op as far as project targets are concerned, one macos CI seems to have segfaulted in one of the tests:

```
12/14 Test #12: yaml_fuzz_to_json ................***Exception: SegFault  0.44 sec
Testing x100, seed: 4064242247
```

mjp41:
I think this is an open issue:
https://github.com/microsoft/Trieste/issues/125
that we have detected in Fuzz before.

PR #164, opened 2025-07-01
Drop requirement that language names match in parsed Trieste files

EliasC:
Parsing a Trieste file used to fail if the language names did not match. This PR insteads logs a debug message if this is the case, but does not require it. The use case for this is targeting an external Trieste tool like vbcc. There is also some overdue formatting fixes.

### DIFF
--- a/include/trieste/driver.h
+++ b/include/trieste/driver.h
@@ -66,6 +66,10 @@ namespace trieste
       std::filesystem::path output;
       build->add_option("-o,--output", output, "Output path.");
 
+      std::string language_name = reader.language_name();
+      build->add_option(
+        "-n,--language_name", language_name, "Language name to use for the output file.");
+
       std::filesystem::path dump_passes;
       build->add_option(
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
Looks fine to me.  I guess I wonder if we should add a feature to change the language output name?  Then this would be an intentional output targeting thing?

matajoh (dismissed):
Just a small thing, but there are a lot of superfluous whitespace changes in this PR. It would be good to keep the changes to just the actual functional edit for clarity.

EliasC:
@matajoh formatting changes reverted! This was me compensating for forgetting to run the formatter in a previous PR.

EliasC:
> @mjp41: Looks fine to me. I guess I wonder if we should add a feature to change the language output name? Then this would be an intentional output targeting thing?

Added specifying language name as an option. We could revert the change and now require names to match, but I don't know that this would help anyone.

mjp41 (approved):
LGTM

PR #168, opened 2025-09-15
Off-by-one bug in Fuzzer

matajoh:
Fixes an off-by-one bug in the Fuzzer that was causing the last pass to never be tested. Fixing this bug then exposed a bug in the final pass of the yaml parser, which this PR also fixes.


EliasC (approved):
LGTM

PR #170, opened 2025-09-25
JSON unicode escaping

matajoh:
`json::escape` was escaping all Unicode characters in strings by default. However, as a result it was also inadventantly encoding UTF16 characters, which is not supported by the codec. Furthermore, it was double-encoding the backslash, which resulted in subsequent parse errors.

The fix is that `json::encode` no longer automatically escapes UTF8 characters. If that is the desired behaviour, there is a new `json::escape_unicode` which escapes all UTF8 characters and (correctly) encodes a Unicode BAD value (`0xFFFD`) if a UTF16 value is encountered.


mjp41 (approved):
LGTM

PR #173, opened 2025-10-08
Adding synthetic origins for readers and thread-local log level

matajoh:
This PR addresses two small issues and makes a significant change to logging:

1. While sources could have synthetic origins, there was no way to specify this when creating a synthetic origin for a reader. This fixes that issue.
2. The setting of log level from a string was case sensitive. This change makes it case insensitive to make the API slightly easier to use.
3. It is sometimes desirable to have different logging behaviour depending on the thread, but previously this was not possible due to the log level being a global property. This change allows more granular behaviour by making log level thread-local. As part of this change, it adds a new RAII `LocalLogLevel` object which allows the log level to change for a particular block of code.


include/trieste/logging.h
```diff
@@ -63,11 +65,19 @@ namespace trieste::logging
...
     };
 
-    // Used to set which level of message should be reported.
-    inline LogLevel report_level{LogLevel::Output};
+    // Used to set which default level of message should be reported.
+    inline LogLevel default_report_level{LogLevel::Uninitialized};
+
+    static detail::LogLevel& thread_local_report_level()
+    {
+      static thread_local detail::LogLevel report_level = default_report_level == LogLevel::Uninitialized ? LogLevel::Output : default_report_level;
```
mjp41, line 78: So this is a non-trivial initialiser, this will result in a branch on every access to the `report_level` to check if it has been initialised, because C++ hates you. This puts us on two branch for every log message that is not printed.  If we detect this case in `start`, then for the not printed messages, we can get down to one branch except for the first message.

mjp41 (approved):
LGTM, minor code suggestion, but happy either way.  My suggestion probably doesn't quite meet clangformat requirements.

include/trieste/logging.h
```diff
@@ -124,6 +135,27 @@ namespace trieste::logging
...
+        }
+        else
+        {
+          detail::thread_local_report_level() = detail::LogLevel::Output;
+          if (level > detail::LogLevel::Output)
+          {
+            return;
+          }
+        }
+      }
```
mjp41, line 157: Perhaps clearer as
```suggestion
      if (
        detail::thread_local_report_level() == detail::LogLevel::Uninitialized)
      {
        detail::thread_local_report_level() = 
          (detail::default_report_level != detail::LogLevel::Uninitialized) ?
            detail::default_report_level :
            detail::LogLevel::Output;

        if (level > detail::thread_local_report_level())
        {
          return;
        }
      }
```

PR #174, opened 2025-10-08
Decoupled utility functions from json/reader.cc and yaml/event_writer.cc

matteB10:
Moved some utility functions from json/reader.cc and yaml/event_writer.cc to new files in their respective directories to improve cohesion and obtain more accurate results when analysing code coverage.


matajoh (changes requested):
This all makes sense, and matches what we discussed. I have a small change: instead of `utils.cc` can you name the files `json.cc` and `yaml.cc`?

PR #179, opened 2025-12-03
Using symbol table bindings in random tree generation

matteB10:
This PR introduces a new option `--gen_bound` that uses bound names during fuzzing.  
This option is enabled by default and leverages symbol tables to find already bound locations when generating a node that:

1. Serves as a binding key in a symbol table within the current scope,
2. Does **not** bind a node defined with `flag::shadowing`, and
3. Is not the binding site itself.

For 1. Bound locations are only used when generating nodes of the same token type as the binding key, and we can find a bound symbol of this type in the current scope. E.g., if we rewrite an `Ident` to a `Ref`, but the symbol table only contains bindings with keys of type `Ident`, the generated `Ref` will get a fresh location. 

For 2. Nodes that bind elements defined with `flag::shadowing` would cause a "conflicting definitions" error when building symbol tables, so we avoid generating trees with such conflicts by checking for this flag.

For 3. If we have something like `Assign <<= Ident * Expr[Ident]` and we are generating an `Ident` as the first child of an `Assign`, we generate a fresh location for that `Ident` since this is the binding site.

Following Matt P's suggestion, if we have `n` bound symbols to choose from, we randomly select one from `n+1` options to occasionally (though rarely) use fresh locations even when the criteria above are met.

### DIFF
--- a/parsers/yaml/to_json.cc
+++ b/parsers/yaml/to_json.cc
@@ -36,6 +36,7 @@ namespace
     | (FlowSequence <<= wf_strings_flow_tokens++)
     | (FlowMappingItem <<= (Key >>= wf_strings_flow_tokens) * (Value >>= wf_strings_flow_tokens))
     | (MappingItem <<= (Key >>= wf_strings_tokens) * (Value >>= wf_strings_tokens))
+    | (TagValue <<= TagPrefix * TagName * (Value >>= wf_strings_tokens))
     ;
   // clang-format on
 
@@ -44,12 +45,13 @@ namespace
 
   // clang-format off
   inline const auto wf_lookup =
-    wf_strings
+    wf_strings 
     | (Document <<= Directives * DocumentStart * (Value >>= wf_lookup_tokens) * DocumentEnd)
     | (Sequence <<= wf_lookup_tokens++)
     | (FlowSequence <<= wf_lookup_flow_tokens++)
     | (FlowMappingItem <<= (Key >>= wf_lookup_flow_tokens) * (Value >>= wf_lookup_flow_tokens))
     | (MappingItem <<= (Key >>= wf_lookup_tokens) * (Value >>= wf_lookup_tokens))
+    | (TagValue <<= TagPrefix * TagName * (Value >>= wf_lookup_tokens))
     ;
   // clang-format on
 


include/trieste/wf.h
```diff
@@ -155,13 +161,82 @@ namespace trieste
...
+              n->type() & flag::lookup &&
+              binding_keys[n->type()].first == type);
+          });
+        }
+        return symbols;
+      }
 
-      Location location(Node n)
+      // Returns true if we should generate a bound variable
+      bool gen_bound(Node parent, Token child_type)
```
EliasC: `gen_bound` sounds like it generates a bound term. Maybe `should_gen_bound`?

include/trieste/wf.h
```diff
@@ -155,13 +161,82 @@ namespace trieste
...
+          auto key_field_index = binding_keys[parent->type()].second;
+          auto key_field_type = binding_keys[parent->type()].first;
+          auto current_index = parent->size() - 2; // Child is last pushed node
+          if (key_field_type == child_type)
+            // Generate a bound variable if we are not at the binding site
+            return key_field_index != current_index;
+          else
+            return false; 
+        }
+        else // Not at binding site
```
EliasC: If you flip the previous condition you can put this first for an early return

include/trieste/wf.h
```diff
@@ -155,13 +161,82 @@ namespace trieste
...
+          auto type = child->type();
+          Nodes symbols = get_symbols(type, parent->scope());
+
+          // Ensure we're not always use bindings when they exist by adding +1
+          auto rand_symbol = static_cast<size_t>(next() % (symbols.size() + 1));
+          // If we have a bound symbol to use, we're not at the binding site,
+          // and the parent type allows shadowing, use a bound symbol
+          if (
+            rand_symbol < symbols.size() && gen_bound(parent, type) &&
+            !(parent->type() & flag::shadowing))
```
EliasC: Shouldn't this last thing be checked inside `gen_bound`? The `parent` here is the parent of the node whose location we're generating, but the shadowing check should only come into play when we happen to be generating a bound variable where a new (shadowing) binding should go.

PR #188, opened 2026-02-11
Allow fuzzing passes in sequence

EliasC:
This PR adds support for fuzzing passes in sequence rather than one at a time. Running a driver with `test --sequence` will generate the requested number of ASTs and pass any survivors on to the next pass, and so on until there are no survivors or until we reach the last pass. The use case for this is for example generating programs for the type checking phase so that the backend can be tested on type checked programs (which survived type checking).

The PR includes a major refactoring of the fuzzer which makes it less of a monolith. There is also options for collecting and presenting size information for generated ASTs.

### DIFF
--- a/include/trieste/driver.h
+++ b/include/trieste/driver.h
@@ -123,7 +123,7 @@ namespace trieste
       test->add_flag("--size_stats", test_size_stats, "Collect size statistics for ASTs (defaults to log level Info)");
 
       test->callback([&]() {
-        if (test_size_stats && logging::detail::default_report_level == logging::detail::LogLevel::Uninitialized)
+        if (test_size_stats && log_level.empty())
         {
           logging::set_log_level_from_string("Info");
         }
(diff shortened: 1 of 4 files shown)


include/trieste/ast.h
```diff
@@ -900,6 +900,34 @@ namespace trieste
...
+      {
+        size_t max_height = 0;
+        for (Node& child : children)
+        {
+          size_t child_height = child->tree_height();
+          max_height = std::max(max_height, child_height + 1);
+        }
+        return max_height;
+      }
+    }
```
mjp41: Could this be rewritten with traverse, so we don't blow the stack on large trees.

```C++
size_t curr_depth = 0;
size_t max_depth = 0;

traverse([&](Node&) {
  curr_depth++;
  if (curr_depth > max_depth)
    max_depth = curr_depth;
  return true;
},
{
  curr_depth--;
});
```

PR #198, opened 2026-04-08
Testsuite: reject crashes in golden files and support multiple suites

mjp41:
Add a validation gate (cmake/validate_golden.cmake) that runs between test execution and golden-file copying during update-dump. It rejects non-numeric exit codes (e.g. "Segmentation fault", "Aborted") so that crashes are never silently accepted as expected output. Collection .cmake files can optionally set TESTSUITE_VALIDATOR to a custom CMake script for additional checks.

Restructure targets so testsuite() can be called multiple times in a project. Each call now creates per-suite targets (e.g. infix-update-dump, infix-update-dump-clean) and the global update-dump / update-dump-clean targets aggregate all suites via add_dependencies. Target names use '-' separators for MSVC generator compatibility. Duplicate testsuite() calls with the same name are caught with a clear error.

Add a shrubbery testsuite (with two example tests) to exercise the multi-suite support alongside the existing infix testsuite.


EliasC (approved):
LGTM! Can also confirm that it solves the issue I was seeing with multiple test suites.

PR #199, opened 2026-04-20
Add location information to .trieste AST dump format

mjp41:
Each AST node with a non-empty source origin now emits its filename, byte offset, and length in the serialized `.trieste` format.

### Format

The grammar for the location annotation after each token name:

```
    location  ::= filename? '|' pos '|' (content | len)
                | content
                | ε

    filename  ::= netstring
    content   ::= netstring
    pos, len  ::= NUM
    netstring ::= NUM ':' <NUM bytes>
```

- **Full location**: `10:input.json|3|4:defg` — filename netstring, `|pos|`, content netstring
- **Elided filename**: `|3|4:defg` — filename omitted when it matches the parent's origin
- **Non-print node**: `10:input.json|0|120` — bare length instead of content netstring
- **Old format** (backward compat): `5:hello` — bare content netstring, no location
- **Bare token**: no annotation

The `|` delimiter separates location fields; `:` appears only inside netstrings. The format is unambiguous without token metadata: after `|pos|`, a netstring is attempted for inline content; if no netstring follows, a bare number gives the span length.

### Location reconstruction

`build_ast()` reconstructs `Location` objects with the original filename and byte offset. Print node sources use `SourceDef::synthetic_at_offset()`, which stores an offset so that `view()` indexes correctly while `pos` carries the original file offset. For offset sources, `linecol()` returns `{0, pos}` as a best-effort approximation; accurate line:col would require serializing line:col in the dump format.

### Other changes

- Filename elision in `str()` uses **origin string comparison** (not pointer identity) so that elision works correctly after deserialization, where each node gets a distinct `SourceDef`.
- `SourceDef` gains an `offset_` field and `synthetic_at_offset()` factory. `Location::view()` uses `pos - offset` to index into the buffer.
- The `File` node created by the parser now spans the whole source file rather than carrying the filename as its location text.
- `build_ast()` validates netstring lengths against remaining input and aborts on malformed location annotations.
- Old-format `.trieste` files without location annotations parse identically to before.
- `.gitattributes` forces LF line endings on test inputs and golden files for consistent byte offsets across platforms.


sylvanc (approved):

EliasC (approved):
LGTM! Am I reading this right that the old format can still be picked up by the Trieste parser?

include/trieste/ast.h
```diff
@@ -746,15 +746,51 @@ namespace trieste
...
+        else
+        {
+          if (node->type_ & flag::print)
+          {
+            auto content = loc.view();
+            out << " " << content.size() << ":" << content;
+          }
+
+          origin_stack.push_back(
+            origin_stack.empty() ? &no_origin : origin_stack.back());
```
matajoh: I'm not sure I follow the logic here. If a location has no source, you duplicate the top of the stack (or add the `no_origin` sentinel)? Wouldn't it be cleaner to always push `no_origin`?

Otherwise, isn't this essentially the value already stored in `effective`?

include/trieste/regex.h
```diff
@@ -860,32 +860,127 @@ namespace trieste
...
+    static const auto hd =
+      TRegex("[[:space:]]*\\([[:space:]]*([^[:space:]\\(\\)]*)");
+    static const auto st = TRegex("[[:space:]]*\\{[^\\}]*\\}");
+    static const auto ns = TRegex("[[:space:]]*([[:digit:]]+):");
+    static const auto colon_num = TRegex("[[:space:]]*:([[:digit:]]+)");
+    static const auto num = TRegex("[[:space:]]*([[:digit:]]+)");
+    static const auto colon = TRegex(":");
+    static const auto tl = TRegex("[[:space:]]*\\)");
+
+    TRegexMatch m(2);
```
matajoh: Why the name change? re_match was quite descriptive.

mjp41:
> LGTM! Am I reading this right that the old format can still be picked up by the Trieste parser?

Yes.  But I haven't tested that as much.  Would you depend on that?  Is so, do you have any tests I could add?

include/trieste/source.h
```diff
@@ -60,11 +66,27 @@ namespace trieste
...
+      source->content_at_zero_ = true;
+      return source;
+    }
+
     const std::string& origin() const
     {
       return origin_;
     }
 
+    bool is_content_at_zero() const
```
matajoh: I dislike this design decision. It breaks encapsulation in a weird way (by which I mean the Location needs to understand internal mechanics of source creation in order to use them properly). One can imagine alternatives which would be equivalent, for example a map of positions to strings that can be populated during the pre-traversal, that would allow the Source to still present the same interface to Location (and any other downstream consumers). It is implemented correctly as far as I can tell, but it feels brittle.

samples/shrubbery/testsuite/examples/operators_out/ast.txt
```diff
@@ -2,29 +2,29 @@ shrubbery
...
-            (shrub-terms
-              (shrub-atom 1:b))
-            (shrub-none)
-            (shrub-none))
-          (group
-            (shrub-terms
-              (shrub-atom 1:c))
-            (shrub-none)
-            (shrub-none)))))))
+  (file 19:operators.shrubbery:0:19
```
matajoh: Small formatting nit: could we use a different delimiter for the netstring? I think it would read more cleanly. I always shudder a bit at the same delimiter being used for different levels of parsing.

matajoh (commented):
Just a few design questions, implementation otherwise appears correct and functional.

EliasC:
> Would you depend on that? Is so, do you have any tests I could add?

No, I have a bachelor student who will be generating Trieste trees, but we can see it as an upcoming test.

include/trieste/ast.h
mjp41: Sorry, I hadn't spotted that when reviewing the generated code.  Simplified the code as you suggested.

include/trieste/regex.h
mjp41: Reverted

include/trieste/source.h
mjp41: As we discussed, I have changed this to now store an offset inside source, and subtract that from any supplied offset.

samples/shrubbery/testsuite/examples/operators_out/ast.txt
mjp41: Changed to use | for field separators and : is for the netstring.  This is way clearer, thanks.

sylvanc (approved):

PR #204, opened 2026-05-26
Compute `pass_through` correctly for sequences

EliasC:
This PR fixes an issue where the `FastPattern` computation would always consider a sequence as having no passthrough. This meant that a pattern like `In(Foo) * Start * T(Bar)` would not end up with `Bar` in its start set (instead it would apply to all tokens in `Foo`, which is less efficient). Passthrough also needs to be checked during pattern compilation as a sequence of only passthrough patterns (which may consume zero nodes) that has a non-empty start set otherwise is only applied when the next token is in that start set. 

Because this only improves efficiency of pattern compilation, there is no regression to test. However, the Shrubbery example contains a sequence pattern with a non-empty start set but that is passthrough:
```c++
In(Group) * Start * (!T(Block, Alt))++[Atom] * ~T(Block)[Block] * ~T(Alt)[Alt] * End
```

### DIFF
--- a/include/trieste/pass.h
+++ b/include/trieste/pass.h
@@ -217,14 +217,16 @@ namespace trieste
       {
         const auto& starts = rule.first.value.get_starts();
         const auto& parents = rule.first.value.get_parents();
+        const bool is_pass_through = rule.first.value.is_pass_through();
 
         //  This is used to add a rule under a specific parent, or to the
         //  default.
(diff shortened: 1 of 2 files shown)


EliasC:
This bug was discovered during a bachelor thesis project.

mjp41 (approved):
LGTM.  Nice work

PR #205, opened 2026-06-09
Abort generation when running out of retries

EliasC:
The current behaviour in the fuzzer is to blindly keep generating without checking for hash uniqueness when the retry budget runs out. This PR instead makes the fuzzer give up on that particular seed and move on, avoiding reusing the same tree over and over.


matteB10 (commented):
Change name to gen_fresh_ast since an ast is not always generated.

include/trieste/fuzzer.h
```diff
@@ -221,6 +224,13 @@ namespace trieste
         hash = ast->hash();
         context.retries++;
       }
+
+      if (context.ast_hashes.find(hash) != context.ast_hashes.end())
```
matteB10: Maybe don't recalculate find

EliasC:
@matteB10 Addressed your comments now!

matteB10 (approved):

mjp41 (approved):
LGTM

PR #206, opened 2026-06-10
Add support for token weights during fuzzing

EliasC:
This PR adds support for supplying token weights to the fuzzer. The weights are supplied as a map from tokens to integers and unmapped tokens defaults to 1. The driver CLI has been extended:
```bash
$ ./infix_trieste test --weights infix-int 2 infix-output 0
# prefer integers, exclude output statements
```
Note that setting a weight to zero may cause generation to abort with an exception if there is no viable option for a choice. I thought about checking for that up front (either checking for paths in the WF spec that lead to dead choices, or propagating dead choices upwards so that the path cannot happen), but I think the benefit doesn't outweigh the benefits in this case.

A possible future thing to add would be an automatic tuning of the weights based on the patterns of a pass: if the patterns have `In(Foo)` and `T(Bar)` that suggests that we should increase the weights of `Foo` and `Bar`. Preliminary experiments suggests that this reduces the number of trivial trees (trees that do not match a single rule).

### DIFF
--- a/include/trieste/wf.h
+++ b/include/trieste/wf.h
@@ -495,7 +495,11 @@ namespace trieste
         for (i = 0; i < min_len; ++i)
           choice.gen(g, depth, node);
 
-        if (depth >= g.target_depth)
+        auto weight_sum = 0;
+        for (const auto& t : choice.types)
+          weight_sum += g.weight_for(t);
+
+        if (depth >= g.target_depth || weight_sum == 0)
         {
           return;
         }
--- a/include/trieste/wf.h
+++ b/include/trieste/wf.h
@@ -497,9 +497,11 @@ namespace trieste
         for (i = 0; i < min_len; ++i)
           choice.gen(g, depth, node);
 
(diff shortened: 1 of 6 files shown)


include/trieste/wf.h
```diff
@@ -80,15 +82,96 @@ namespace trieste
...
+            std::vector<double> offsets;
+            offsets.reserve(tokens.size());
+            std::transform(
+              tokens.begin(),
+              tokens.end(),
+              std::back_inserter(offsets),
+              [&](const Token& t) {
+                return static_cast<double>(weight_for(t));
+              });
+            return choose_weighted(tokens, std::move(offsets), parent, depth);
```
mjp41: Probably doesn't mater, but I wonder if some of the calculation of weights for wf entry should be done once.  My quick reading of this implies you calculate the cumulative distribution each time you select an element?

EliasC: Good catch! Adressed in new (force pushed) version

PR #207, opened 2026-06-15
Moved infix::calculate() from writers.cc, fixed problem with Ref lookups 

matteB10:
1. Moved the `Rewriter` `infix::calculate` into its own file, calculate.cc to decouple and keep code coverage measures separate.
2. Updated the `exists` and `can_replace` action functions. We discovered that the old `can_replace` failed to find any replaceable identifers during fuzzing, including when using --gen-bound to generate variables with symbol‑table bindings. The issue turned out to be that the rule called `can_replace` on the `Ref` node instead of its `Ident` child. Since generated `Ref` nodes don’t always share the same location as their `Ident`, the rule never matched. The updated version fixes this by applying the check directly to the `Ident` node.


samples/infix/calculate.cc
```diff
@@ -0,0 +1,200 @@
...
+      }};
+  }
+
+  // clang-format off
+  const auto wf_to_file =
+    infix::wf
+    | (Top <<= File)
+    | (File <<= Path * Calculation)
+    ;
+}
```
mjp41: Duplicated?

mjp41 (approved):
LGTM, can you fix the copilot comments please.

PR #210, opened 2026-07-31
Fix fast-pattern indexing for negated patterns

mjp41:
The pattern `!Pat`, ensures the current cursor does not match `Pat`, and then skips forward one token.  This is different to `--Pat`, which doesn't advance the cursor.

This fixes the FastPattern for !Pat, which treated it as `--Pat`.  We consider `!Pat` to consume an arbitrary first token.


mjp41:
@eliasc added you as a reviewer as an FYI.  I'll merge before you are back.

PR #211, opened 2026-08-05
Redesign CMake testsuite around named nodes

mjp41:
Replace the legacy collection runner with explicit named DAG nodes, golden and artifact declarations, dependency-aware verification and updates, and shared node execution. Migrate samples, add focused integration coverage, and document the new API and migration path.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


mjp41:
@qhuiw this is my new approach to the testsuite in Trieste, I think this means we can move Verona-BC back to using it, rather than rolling its own slightly different version.  The file testsuite.md explains its usage.

PR #213, opened 2026-09-16
Reject colliding explicit WF field names

mjp41:
- track whether WF field names are explicit or implicit
- reject duplicate effective names when either field uses explicit `>>=` naming
- preserve valid repeated implicit fields such as `LocalId * LocalId`
- add focused regression coverage for construction, append, diagnostics, and binding preservation


mjp41:
This is based on a bug in verona-bc that @qhuiw discovered where duplicate fields occurred due to nested definitions.  In https://github.com/sylvanc/verona-bc/pull/84/changes/924d982bbac2add71830dafe390186caaefe2847 both wfSrc and wfLit define the same field.

EliasC:
LGTM!

EliasC (approved):

